PR #24, opened 2014-03-07
add destroy method

calvinmetcalf:
based on discussion at maxogden/level.js#29 a standardized destroy method on the prototype is a lot more portable then the current thing leveldown has on the constructor, especially as it's easy to call a constructor's destroy method from a prototype method then the other way around.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ Remember that each of these methods, if you implement them, will receive exactly
 
 ### AbstractLevelDOWN(location)
 ### AbstractLevelDOWN#_open(options, callback)
+### AbstractLevelDOWN#_destroy(options, callback)
 ### AbstractLevelDOWN#_close(callback)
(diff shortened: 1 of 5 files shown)


mhsjlw:
Yeah, :+1:

vweevers:
I'm having difficulty pinpointing when and where the decision was made, but methods like `destroy` are `leveldown`-specific and cannot be supported in all implementations. As such, they don't belong in `abstract-leveldown`.

IMO a more generic solution would be a method like `clear`, which would empty a db *while in use*. This can have a default implementation: iterating and deleting all keys. Different implementations can optimize this in their own way.

Such a method would differ from `leveldown#destroy`, which also deletes the parent folder and assumes the db is closed (which is why the method is not on the prototype).

@ralphtheninja I vote for closing this and #42 and to open a new issue for `clear`. WDYT?

ralphtheninja:
@vweevers SGTM!

PR #41, opened 2014-09-18
Implement `collectEntries` as trampoline.

bigeasy:
In pure-JavaScript implementations that aggressively cache records certain test will blow the stack if `process.nextTick` is used. The warning issued suggests the use of `setImmediate` however, not all platforms may support `setImmediate`.

This implementation of `collectEntries` uses a trampoline so that invocations of `iterator.next` that do not result in an asynchronous call will not increase the depth of the stack. This eliminates recursion and prevents stack overflow.

### DIFF
--- a/testCommon.js
+++ b/testCommon.js
@@ -51,6 +51,11 @@ var dbidx = 0
   , collectEntries = function (iterator, callback) {
       var data = []
         , next = function () {
+            var f = _next()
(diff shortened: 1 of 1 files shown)


vweevers:
Closing because #40 was merged (with `setTimeout` instead of `setImmediate`).

PR #51, opened 2015-01-23
providing compare extensibility

deepak1556:
https://github.com/rvagg/node-leveldown/pull/145

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -218,14 +218,15 @@ AbstractLevelDOWN.prototype.approximateSize = function (start, end, callback) {
 }
 
 AbstractLevelDOWN.prototype.compare = function (keyA, keyB) {
+  if (typeof this._compare != 'function') {
+    throw new Error('_compare() is not defined')
+  }
+
   if (keyA == null || keyB == null) {
     throw new Error('compare() requires valid `keyA`, `keyB` arguments')
   }
 
-  if (typeof this._compare == 'function')
-    return this._compare(keyA, keyB)
-
-  throw new Error('_compare() is not defined')
+  return this._compare(keyA, keyB)
 }
 
 AbstractLevelDOWN.prototype._setupIteratorOptions = function (options) {
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -224,6 +224,8 @@ AbstractLevelDOWN.prototype.compare = function (keyA, keyB) {
 
   if (typeof this._compare == 'function')
     return this._compare(keyA, keyB)
+
+  throw new Error('_compare() is not defined')
 }
 
 AbstractLevelDOWN.prototype._setupIteratorOptions = function (options) {
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -217,6 +217,15 @@ AbstractLevelDOWN.prototype.approximateSize = function (start, end, callback) {
   })
 }
 
+AbstractLevelDOWN.prototype.compare = function (keyA, keyB) {
+  if (keyA == null || keyB == null) {
+    throw new Error('compare() requires valid `keyA`, `keyB` arguments')
+  }
+
+  if (typeof this._compare == 'function')
+    return this._compare(keyA, keyB)
+}
+
 AbstractLevelDOWN.prototype._setupIteratorOptions = function (options) {
   var self = this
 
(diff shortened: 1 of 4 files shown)


abstract-leveldown.js
```diff
@@ -190,6 +190,15 @@ AbstractLevelDOWN.prototype.approximateSize = function (start, end, callback) {
   })
 }
 
+AbstractLevelDOWN.prototype.compare = function (keyA, keyB) {
+  if (keyA == null || keyB == null) {
+    throw new Error('compare() requires valid `keyA`, `keyB` arguments')
+  }
+
+  if (typeof this._compare == 'function')
```
juliangruber: maybe we should throw if `this._compare` isn't defined?

juliangruber:
otherwise, lvgtm!

abstract-leveldown.js
deepak1556: yes! thanks :)

juliangruber:
just a little last change, i'd change the order and throw on missing `._compare` first. any other opinions?

watson:
@juliangruber I agree - it's easier to read IMHO if the edge case (the throw) is inside the if-statement

watson:
Also while we're add the nitty bitty details (sorry... can... not... help... it...), it's odd that the first one-liner if-statement uses curly braces and the 2nd doesn't #consistency

Looking forward to implement this in mongodown :smiley_cat: :shipit:

rvagg:
is it going to be a problem for some backends that this is not asynchronous? it should be simple for most embedded dbs but not so much for anything that has a network connection in between.

@watson: how would this go on mongodown?

watson:
@rvagg Hmm... you might be right. If I understand the intention of the compare function correctly it's to take two potential keys see how they would be sorted if used as keys in the database? In that case I don't know of way to "ask" MongoDB about this (e.g. to get access to its internal `$lt`/`$gt` functions?). In case they did expose this in some way then yeah it had to be implemented asynchronously. That being said, I'm almost certain that it uses JavaScript to compare strings, so it would be trivial to implement (fake) in a synchronous manner directly in MongoDOWN - but this is all an educated guess of cause...

ralphtheninja:
When https://github.com/deepak1556/abstract-leveldown/pull/1 has been merged we can merge this and release a new minor.

ralphtheninja:
lgtm 

@juliangruber anything you're missing or something that should be adjusted?

abstract-leveldown.js
```diff
@@ -217,6 +217,16 @@ AbstractLevelDOWN.prototype.approximateSize = function (start, end, callback) {
   })
 }
 
+AbstractLevelDOWN.prototype.compare = function (keyA, keyB) {
+  if (typeof this._compare != 'function')
+    throw new Error('_compare() is not defined')
+
+  if (keyA == null || keyB == null)
```
juliangruber: shouldn't `null` values be valid?

ralphtheninja: It can't hurt. What should the convention be? `null` is always less than non null?

ralphtheninja: And I guess `==` here is used to catch both `null` and `undefined`?

juliangruber: it's for the actual backend to decide that, for example a json encoding might sort it differently than a utf8 encoding

ralphtheninja: > it's for the actual backend to decide that, for example a json encoding might sort it differently than a utf8 encoding

Makes sense. This `compare()` method will not do much then :)

ralphtheninja: We do all these `this._checkKey()` calls in all methods which checks for `null` and `undefined` and they return an error.

Given that, then maybe this method should also produce an error for an invalid key? Or maybe just do something like:

```js
if (this._checkKey(keyA) || this._checkKey(keyB)) {
  // ..
}
```

vweevers:
Closing because it's stale. A new PR would be welcome, with the following changes (accounting for features and logic that have since been added to `abstract-leveldown`):

- Do `key = this._serializeKey(key)`
- Do `this._checkKey(key)`
- Add an options argument (so that one can specify encoding options for `encoding-down`)
- Add `compare` to the manifest (because not all implementations can support it)
- Allow null and undefined keys, because those are valid as range options (with a `bytewise/charwise` encoding) and `compare()` is more related to range options than keys.

PR #58, opened 2015-05-07
add sync put tests

juliangruber:
This adds 2 sync put tests, trying to recreate the failure from https://github.com/Level/leveldown/issues/157. On leveldown@master, this does __not__ cause a segfault for me. A `.put()` after the db is fully closed however does.

Once this is merged, we can add it to backends supporting `{ sync }`.

### DIFF
--- a/abstract/put-test.js
+++ b/abstract/put-test.js
@@ -78,6 +78,28 @@ module.exports.put = function (test) {
 
 }
 
+module.exports.sync = function (test) {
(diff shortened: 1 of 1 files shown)


juliangruber:
if we decide this is the righ place to add such tests, i'll add more sync tests for methods supporting the option

ralphtheninja:
:+1:

juliangruber:
2.2.0

Issue #59, opened 2015-05-12
using empty location + open causes weird results


ralphtheninja:
If I do the following:

```js
var leveldown = require('leveldown')
var db = leveldown('')
db.open(console.log.bind(console))
```

I get the following output:

```
[Error: IO error: /LOCK: Permission denied]
```

Which is kind of odd. I propose that we make `AbstractLevelDOWN` throw an error if a zero length string is given, instead of finding out in `open()` that we couldn't open that location. We already know it will fail and might as well error in the constructor.

juliangruber:
Agreed. `.` and even `/` should be valid locations, but not the empty string. `if (!location) ...` should be fine

PR #60, opened 2015-05-12
Empty location

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/59

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -8,10 +8,7 @@ function AbstractLevelDOWN (location) {
   if (!arguments.length || location === undefined)
     throw new Error('constructor requires at least a location argument')
 
-  if (typeof location != 'string')
-    throw new Error('constructor requires a location string argument')
-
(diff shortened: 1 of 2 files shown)


abstract-leveldown.js
```diff
@@ -11,6 +11,9 @@ function AbstractLevelDOWN (location) {
   if (typeof location != 'string')
     throw new Error('constructor requires a location string argument')
```
juliangruber: i'd merge the two into something like

```js
if (typeof location != 'string' || !location.length)
  throw new Error('constructor requires a non empty location string')
```

that should be enough information

juliangruber:
otherwise lgtm :)

abstract-leveldown.js
ralphtheninja: Good call. I'll update.

ralphtheninja:
```
$ npm publish
+ abstract-leveldown@2.2.1
```

PR #62, opened 2015-05-13
use t.equal instead of t.ok(a === b)

juliangruber:


ralphtheninja:
:+1: `equal` is much better because it gives more information when it fails.

juliangruber:
hurray :)

PR #64, opened 2015-05-17
add isLevelDOWN() function

ralphtheninja:
This adds the function `isLevelDOWN()` which I extracted from @substack 's commit https://github.com/substack/node-levelup/commit/cff0ae9fdf527569d20d48c73804dd8dafea2b59 (I'm cherry picking those commits to a new branch to make a new PR to levelup).

It basically makes sure the public api in `AbstractLevelDOWN` (except for the `approximateSize` method) is present in the given db object.

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -242,18 +242,4 @@ AbstractLevelDOWN.prototype._checkKey = function (obj, type) {
     return new Error(type + ' cannot be an empty String')
 }
 
-function isLevelDOWN (db) {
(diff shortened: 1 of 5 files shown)


ralphtheninja:
Actually. Hold off with this PR. We could put that function in a separate file instead and add `index.js` that exports everything.

juliangruber:
+1

ralphtheninja:
```
$ npm publish
+ abstract-leveldown@2.3.0
```

Issue #63, opened 2015-05-13
make location optional


juliangruber:
The [recent change](https://github.com/Level/abstract-leveldown/commit/e2547f393d680da861cf0c2fed0e52184b9277dd) to validate that a location isn't an empty string was actually a breaking change, as modules like [memdown here](https://github.com/Level/memdown/blob/master/memdown.js#L124) may pass an empty string.

I'm for making location optional as it has shown that enough backends don't need one.

juliangruber:
we could even go one step further and say we pass `meta` or `options` to the underlying backend, as `location` is coupled to certain implementations, and others might need others. A dynamo backend for example needs `.key` and `.secret`.

ralphtheninja:
> I'm for making location optional as it has shown that enough backends don't need one.

:+1:

ralphtheninja:
This will simplify a lot. Taking a look :)

ralphtheninja:
Old one which wasn't added to the backlog. And already assigned to me. How practical :)

PR #65, opened 2015-05-18
Optional options

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/63


juliangruber:
+1

ralphtheninja:
I'll rebase this.

juliangruber:
this is a new major, right?

ralphtheninja:
@juliangruber Yes this should be a new major since we have removed `.location` and replaced it with `.options`.

juliangruber:
ok, then when do you think we should release this?

ralphtheninja:
Today? :)

ralphtheninja:
Let me merge master into this branch first. We can have this PR hanging for a while for others to comment.

ralphtheninja:
@juliangruber I don't remember why this was closed. Not saying it should be re-opened, just wondering about the reasoning behind it. If we still should get rid of `.location` then make a new PR instead.

PR #78, opened 2016-01-05
Only show build status for the master branch

watson:
Since merging #77, the tests now run for all branches (which is good!). But the build status badge in the README.md showed the build status for the latest build - no matter the branch. So if a pull request failed its tests the main badge would show as failed on both npm and github.

This commit ensures that only the state of the master branch is reflected in the build status badge.

This commit also updates the badge URL to the latest recommend URL by Travis.


juliangruber:
Thanks!

PR #81, opened 2016-02-10
iterator-test: don't coerce values in the browser

timkuijsten:
Currently trying to get native JS type support in [level-js](maxogden/level.js#46). This little patch is required to make all abstract tests pass in the browser.


vweevers:
No longer necessary; since #140 we use string values.

Issue #84, opened 2016-02-14
new encoding schemes for backends that support JS types


timkuijsten:
For backends like IndexedDB that support native JS types, should we introduce something like `none`, `js` or `native` as a valid encoding for keyEncoding and valueEncoding?

related to #83

juliangruber:
you can already write a custom encoding, a la

```js
var noop = function(){};
var none = { encode: noop, decode: noop };
db({ valueEncoding: none, keyEncoding: none });
```

iirc that should work, but haven't tried it

vweevers:
This is supported, but note there's two layers to this.

1. Encoding. The user of `level(up)` must use the [`none` aka `id` encoding](https://github.com/Level/codec/blob/v7.1.0/lib/encodings.js#L41). E.g. `valueEncoding: 'id'`.
2. Serialization. The backend must implement `_serialize*` as identity functions:

```js
down.prototype._serializeKey = function (key){ return key }
down.prototype._serializeValue = function (value){ return value }
```

It's very likely we'll [make this the default behavior](https://github.com/Level/abstract-leveldown/issues/121#issuecomment-328336205). So in the future, only the backends that *need* to (because they can only store a subset of types - e.g. buffers and strings in `leveldown`) will have to implement `_serialize*`.

Issue #83, opened 2016-02-14
binary encoding vs. asBuffer


timkuijsten:
I'm wondering what the exact purpose of the asBuffer option is (and it's cousins keyAsBuffer and valueAsBuffer). To me it feels redundant since one can use keyEncoding `binary` and/or valueEncoding `binary`.

I wonder what to do when using IndexedDB as a backend which has support for native JS types as values and supports some different types for keys (maxogden/level.js#48). I know in level.js they even have a `raw` option. I can't really tell the difference between encoding, asBuffer and raw.

/cc @nolanlawson

vweevers:
> I'm wondering what the exact purpose of the asBuffer option is (and it's cousins keyAsBuffer and valueAsBuffer). 

The `asBuffer` option tells the store to fetch data as a Buffer, and avoid unnecessary conversion. Stores like `leveldown` have the ability to return either a Buffer or string, both sourced from the same byte array. See [encoding format](https://github.com/Level/codec#encoding-format) for more.

> To me it feels redundant since one can use keyEncoding binary and/or valueEncoding binary.

The `binary` encoding is just one of the encodings (which is user-specified) that result in `asBuffer: true` (which is derived from the encoding).

> I wonder what to do when using IndexedDB as a backend which has support for native JS types as values and supports some different types for keys (maxogden/level.js#48)

TBD, to be discussed elsewhere.

>  I know in level.js they even have a raw option.

IMO the non-standard `raw` option should be removed from `level.js`. And if `asBuffer` is false, `level.js`  should return the value as-is. Then, by using the `identity` encoding, one can achieve the same effect as `raw: true`.

vweevers:
See also Level/level.js#58

PR #85, opened 2016-02-24
Remove stringification of keys and values

juliangruber:
levelup handles massaging keys and values through its encoding layer already. However, if the user passed `encoding: 'id'`, abstract-leveldown would still convert everything to a string or buffer. For those encodings, like level.js and postgres-down that do support more of the native JS values, this gives them the chance of receiving the raw values and encoding them themselves.

It's a breaking change and so if merged will need a `3.0.0` release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,12 @@ See [MemDOWN](https://github.com/Level/memdown/) if you are looking for a comple
 
 Remember that each of these methods, if you implement them, will receive exactly the number and order of arguments described. Optional arguments will be converted to sensible defaults.
 
-### AbstractLevelDOWN(location)
+### AbstractLevelDOWN(location, options)
+
+Available options:
+
+- `toBuffer`: Convert all keys and values to buffers. Default: `true`.
+
 ### AbstractLevelDOWN#_open(options, callback)
 ### AbstractLevelDOWN#_close(callback)
 ### AbstractLevelDOWN#_get(key, options, callback)
--- a/README.md
+++ b/README.md
@@ -89,12 +89,7 @@ See [MemDOWN](https://github.com/Level/memdown/) if you are looking for a comple
 
 Remember that each of these methods, if you implement them, will receive exactly the number and order of arguments described. Optional arguments will be converted to sensible defaults.
 
-### AbstractLevelDOWN(location, options)
-
-Available options:
-
-- `toBuffer`: Convert all keys and values to buffers. Default: `true`.
-
+### AbstractLevelDOWN(location)
 ### AbstractLevelDOWN#_open(options, callback)
 ### AbstractLevelDOWN#_close(callback)
 ### AbstractLevelDOWN#_get(key, options, callback)
@@ -109,6 +104,8 @@ If `batch()` is called without arguments or with only an options object then it 
 By default a `batch()` operation without arguments returns a blank `AbstractChainedBatch` object. The prototype is available on the main exports for you to extend. If you want to implement chainable batch operations then you should extend the `AbstractChaindBatch` and return your object in the `_chainedBatch()` method.
 
 ### AbstractLevelDOWN#_approximateSize(start, end, callback)
+### AbstractLevelDOWN#_serializeKey(key)
+### AbstractLevelDOWN#_serializeValue(value)
 ### AbstractLevelDOWN#_iterator(options)
 
 By default an `iterator()` operation returns a blank `AbstractIterator` object. The prototype is available on the main exports for you to extend. If you want to implement iterator operations then you should extend the `AbstractIterator` and return your object in the `_iterator(options)` method.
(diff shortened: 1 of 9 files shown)


joeybaker:
I never leave a drive-by, :+1:, but @juliangruber asked for 'em, so: +1 from me!

juliangruber:
thanks @joeybaker :heart:

ralphtheninja:
+1

deanlandolt:
Huge :+1:

rvagg:
How could we best communicate this change to downstream users? It's a pretty big one and could open up some problems that won't necessarily be picked up by test suites.

deanlandolt:
Perhaps this could be opt-in? A prototype key an implementation has to set to remove this behavior?

ISTM this would would also help address #83, #84 as well (and a number of other issues across a whole slew of repos, I'm sure).

juliangruber:
or

```js
function DB(){
  AbstractLevelDOWN.call(this, name, {
    toBuffer: false
  })
}
```

rvagg:
That'd actually be a great first step regardless, make it opt-in for now, switch to opt-out and maybe remove in the future.

juliangruber:
check this out

deanlandolt:
LGTM :+1:

But why not `this._db._toBuffer` in `AbstractChainedBatch`?

An alternate design (perhaps a bit more flexible) just came to mind: add a `_serialize` function to `AbstractLevelDOWN` like this:

```js
AbstractLevelDOWN.prototype._serialize = function (input) {
  return this._isBuffer(input)) ? input : String(input)
}
```

Then on `AbstractChainedBatch` you can just add another `_serialize` function that falls through to `this._db._serialize`, but allows you to customize the behavior per chained batch. This would represent the *last* step before data gets pushed down into the store. This would also allow you to, e.g. force values into typed arrays, or whatever a given backing store prefers, w/o having to completely reimplement the rest of the write logic.

Does this sound reasonable?

juliangruber:
this sounds _mega perfect_! i've thought about this for a little and it so makes sense. i'll work on it

juliangruber:
actually, it needs to be slightly more complicated, because serialization depends on whether it's a key or a value. i'll create `_serializeKey` and `_serializeValue`

juliangruber:
ok how does this look?

deanlandolt:
@juliangruber: that's beautiful, man! :+1:

juliangruber:
added a couple missing tests and fixed the bug magnus discovered. mergeable as a minor?

ralphtheninja:
> mergeable as a minor?

+1

juliangruber:
thanks @ralphtheninja and @deanlandolt! @rvagg can i get a +1 from you as well?

deanlandolt:
:+1:

juliangruber:
2.5.0 it is!

nolanlawson:
So it looks like this was indeed a breaking change, because AFAICT from git-bisecting, this commit broke memdown: https://github.com/Level/memdown/issues/55

nolanlawson:
BTW maybe we should start using something like [dont-break](https://github.com/bahmutov/dont-break) in this project, because I feel like we run into the situation a lot where somebody merges something and then it randomly breaks FooDOWN somewhere in the ecosystem.

ralphtheninja:
> BTW maybe we should start using something like dont-break in this project, because I feel like we run into the situation a lot where somebody merges something and then it randomly breaks FooDOWN somewhere in the ecosystem.

+1

juliangruber:
you're right that this actually is a breaking change. Before, `null` and `undefined` would be stored as empty string values, now they get passed through. Those exact tests are failing:  https://github.com/Level/abstract-leveldown/blob/master/abstract/put-get-del-test.js#L140-L141.

I'm working on a pull request to revert this breaking change.

juliangruber:
Fixed in https://github.com/Level/abstract-leveldown/commit/7f514f439c24010e650a7587c308d6d9fd75930d, committed straight to master and release as 2.6.1. The memdown test suite passes with this version, I'll get it upgraded next.

nolanlawson:
Awesome, thank you!! :) So just to check - this changes the on-disk representation of data? I.e. if I created data with <2.5.0, if I then open it in 2.6.0 the data may be changed/corrupted?

juliangruber:
it changes how data nullish values are stored in 2.6.0., 2.6.1 however will behave the same as 2.5.0

nolanlawson:
Commented on this in https://github.com/Level/memdown/pull/57#issuecomment-246856206, but there's something basic I'm not understanding: if I write data in 2.4.1, then read that same data in 2.6.1, is the data changed/corrupted? If so, then 2.5.0 should have been a major version bump, and in addition it would need to offer some guidelines on how to migrate userdata over (or determine if a migration is necessary).

Yes this only applies to nullish data, and yes memdown isn't affected, but this project has [a lot of dependents](https://www.npmjs.com/browse/depended/abstract-leveldown) so it's worth communicating downstream if package authors need to figure out a migration path.

juliangruber:
only 2.6.0 is the only faulty version. 2.6.1 and 2.5.0 have (or should have) full compatibility with anything done with 2.4.1.

nolanlawson:
OK, I misunderstood. Thanks for clarifying!

Issue #87, opened 2016-03-04
AbstractChainedBatch tests should commit and inspect db rather than sniff `_operations` buffer


deanlandolt:
Nothing about `AbstractChainedBatch` requires it to keep operations around in memory (in the `_operations` buffer), but some of its tests sniff this buffer rather than committing these values and reading from the db. The tests are mostly around just verifying the _serialize[Key|Value] behavior, so in this case they would probably be better off hooking `_put` and `_del`. I do have to do all [this nonsense here](https://github.com/ralphtheninja/pgdown/blob/master/test/abstract-leveldown.js#L15-L43)) from outside the tests, but from within we could just hook the `_put` and `_del` methods on the batch directly.

Regardless, these tests should also commit these values and verify what ends up in the db w/ `collectEntries`. The only exception might be `test custom _serialize*`, since what will happen when this hits the db is undefined behavior. But this just suggests to me we should add companion `_deserialize[Key|Value]` methods to get us a hook to reliably verify expected results.

/cc @ralphtheninja @juliangruber

vweevers:
Fixed by #88, except for:

> Regardless, these tests should also commit these values and verify what ends up in the db w/ collectEntries. The only exception might be test custom _serialize*, since what will happen when this hits the db is undefined behavior.

I think what we have is good. Serializing and committing is tested separately, as unit tests should. In addition, after #277, the default serialize functions will be identity functions. The idea behind that is we shouldn't make any assumptions about serialization - including in tests. So indeed, "what will happen when this hits the db is undefined behavior".

PR #88, opened 2016-03-04
Buffer _put/_del inputs in chained batch tests

deanlandolt:
This PR addresses the main problem w/ the chained batch tests mentioned in #87 -- the tests assume an `_operations` buffer is kept in chained batch. This fixes the tests to capture the inputs that _would_ go to an `_operations` buffer, which is the data these tests are trying to introspecting.

/cc @ralphtheninja @juliangruber

### DIFF
--- a/abstract/chained-batch-test.js
+++ b/abstract/chained-batch-test.js
@@ -2,6 +2,28 @@ var db
   , leveldown
   , testCommon
 
+function collectBatchOps (batch) {
+  var _put        = batch._put
+    , _del        = batch._del
+    , _operations = []
+
+  if (typeof _put !== 'function' || typeof _del !== 'function') {
+    return batch._operations
+  }
+
+  batch._put = function (key, value) {
+    _operations.push({ type: 'put', key: key, value: value })
+    return _put.apply(this, arguments)
+  }
+
+  batch._del = function (key) {
+    _operations.push({ type: 'del', key: key })
+    return _del.apply(this, arguments)
+  }
+
+  return _operations
+}
+
 module.exports.setUp = function (_leveldown, test, _testCommon) {
   test('setUp common', _testCommon.setUp)
   test('setUp db', function (t) {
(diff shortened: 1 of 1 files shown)


ralphtheninja:
Other than nitpicky stuff, lgtm!

deanlandolt:
@ralphtheninja any nits other than the braces around the single line if body? Just pushed up a fix for that

ralphtheninja:
> @ralphtheninja any nits other than the braces around the single line if body? Just pushed up a fix for that

Nope. LGTM!

abstract/chained-batch-test.js
```diff
@@ -170,27 +190,39 @@ module.exports.args = function (test) {
...
+      .del({ bar: 'baz' })
+
+    t.deepEqual(ops, [
+        { type: 'put', key: { foo: 'bar' }, value: { beep: 'boop' } }
+      , { type: 'del', key: { bar: 'baz' } }
+    ])
+
+    // remove _serialize* patch
+    db._serializeKey = _serializeKey
+    db._serializeValue = _serializeValue
```
juliangruber: If this test fails before it reaches this line the following tests will fail as well. Can we revert this?

juliangruber:
Otherwise +1!

deanlandolt:
Fair enough -- wasn't sure if this would be an issue or not. But there's an easy way to fix this.

deanlandolt:
@juliangruber what do you think of this approach?

juliangruber:
ah, yeah that's smart!

just two minor things and we can merge:

- can you please remove the comment?
- can you rename `_db` back to `db`? scoping is protecting us from doing anything bad

deanlandolt:
Cool, will do.

deanlandolt:
@juliangruber I pushed up a commit that removes the comment. But actually, a separate var (`_db` or otherwise) is required. Otherwise clobber the `db` in the test's scope and can't delegate to it in `Object.create`. If we replace the `db` var with the new delegate db we'd lose the benefit of delegation entirely.

juliangruber:
alright, thank you! :)

PR #89, opened 2016-03-05
Expose a hook to verify approximate size

deanlandolt:
We need this to enable the `approximateSize` tests in `pgdown` -- `TOAST` compression looks to be around 4x better than `snappy` so the hardcoded `40000` won't cut it. While I was at it I figured it wouldn't hurt to pass in the whole batch to allow implementations to test the approximate size however they see fit.

/cc @ralphtheninja @juliangruber


juliangruber:
.approximateSize is being deprecated from the current level interface, if a backend has it it should be up to the backend to verify this method. Sounds good?

deanlandolt:
Perfect.

juliangruber:
Oh I wasn't thinking we should revert the whole thing, just should add back creating a fresh db  for this test

PR #90, opened 2016-03-09
Allow serialization functions to return buffers

deanlandolt:
This just loosens up the tests to allow _serlialize* functions to force buffers
for string keys or values. The underlying store impl might prefer (or require) buffers.

/cc @juliangruber @ralphtheninja

### DIFF
--- a/abstract/approximate-size-test.js
+++ b/abstract/approximate-size-test.js
@@ -71,8 +71,8 @@ module.exports.args = function (test) {
     t.plan(3)
     var db = leveldown(testCommon.location())
     db._approximateSize = function (start, end, callback) {
-      t.equal(start, '[object Object]')
-      t.equal(end, '[object Object]')
+      t.equal(Buffer.isBuffer(start) ? String(start) : start, '[object Object]')
+      t.equal(Buffer.isBuffer(end) ? String(end) : end, '[object Object]')
       callback()
     }
     db.approximateSize({}, {}, function (err, val) {
--- a/abstract/chained-batch-test.js
+++ b/abstract/chained-batch-test.js
@@ -181,6 +181,10 @@ module.exports.args = function (test) {
     batch
       .put({ foo: 'bar' }, { beep: 'boop' })
       .del({ bar: 'baz' })
+    ops.forEach(function (op) {
+      if (Buffer.isBuffer(op.key)) op.key = String(op.key)
+      if (Buffer.isBuffer(op.value)) op.value = String(op.value)
+    })
     t.deepEqual(ops, [
         { type: 'put', key: '[object Object]', value: '[object Object]' }
       , { type: 'del', key: '[object Object]' }
--- a/abstract/del-test.js
+++ b/abstract/del-test.js
@@ -46,7 +46,7 @@ module.exports.args = function (test) {
     t.plan(2)
     var db = leveldown(testCommon.location())
     db._del = function (key, opts, callback) {
-      t.equal(key, '[object Object]')
+      t.equal(Buffer.isBuffer(key) ? String(key) : key, '[object Object]')
       callback()
     }
     db.del({}, function (err, val) {
--- a/abstract/get-test.js
+++ b/abstract/get-test.js
@@ -46,7 +46,7 @@ module.exports.args = function (test) {
     t.plan(2)
     var db = leveldown(testCommon.location())
     db._get = function (key, opts, callback) {
-      t.equal(key, '[object Object]')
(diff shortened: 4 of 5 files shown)


ralphtheninja:
+1

juliangruber:
publish as a minor?

ralphtheninja:
> publish as a minor?

The tests are in fact part of the api and this is new and backwards compatible functionality, so yes :)

juliangruber:
2.6.0 it is!

ralphtheninja:
Hmm, this broke the tests on leveldown https://travis-ci.org/ralphtheninja/leveldown/jobs/115013099

ralphtheninja:
Actually, the tests on leveldown break at 2.5.0

```
⨯ test serialize object
  not ok 70 should be equivalent
/home/lms/src/level/abstract-leveldown/abstract/chained-batch-test.js:175
    t.equal(batch._operations[0].key.toString(), 'foo')
                                ^

TypeError: Cannot read property 'key' of undefined
    at Test.<anonymous> (/home/lms/src/level/abstract-leveldown/abstract/chained-batch-test.js:175:33)
    at Test.bound [as _cb] (/home/lms/src/level/leveldown/node_modules/tape/lib/test.js:61:32)
    at Test.run (/home/lms/src/level/leveldown/node_modules/tape/lib/test.js:77:10)
    at Test.bound [as run] (/home/lms/src/level/leveldown/node_modules/tape/lib/test.js:61:32)
    at Immediate.next [as _onImmediate] (/home/lms/src/level/leveldown/node_modules/tape/lib/results.js:70:15)
    at processImmediate [as _immediateCallback] (timers.js:383:17)
    ---
      operator: deepEqual
      expected: |-
        [ { key: '[object Object]', type: 'put', value: '[object Object]' }, { key: '[object Object]', type: 'del' } ]
      actual: |-
        []
      at: Test.<anonymous> (/home/lms/src/level/abstract-leveldown/abstract/chained-batch-test.js:164:7)
    ...
✓ test serialize buffer
not ok 71 no plan found
⨯ fail  2
npm ERR! Test failed.  See above for more details.
```

ralphtheninja:
Here's the bisect:

screenshot from 2016-03-11 14 10 58

I think we should be more forceful on squashing PR's since it's difficult to know where something went wrong if some commits are a half finished feature or update.

deanlandolt:
Probably wouldn't be a bad idea to run the leveldown tests in the abstract-leveldown test suite to prevent this kind of thing -- if not in the test script itself, at least as a separate task in travis. Hell, it may not be a bad idea to run the levelup tests, and any other "reference" implementations.

It'd be really nice if there were a bin script to install (w/ dev deps) and run the test suites for a wide swatch of abstract-leveldown implementations, patching the ctors to use the current abstract-leveldown classes.

deanlandolt:
@ralphtheninja the stack you showed for the 2.5.0 breakage should have been fixed by this PR: https://github.com/Level/abstract-leveldown/pull/88 (I did that for pgdown but figured leveldown would need that too -- I was wondering why the usage of `this._operations` wasn't breaking it, but figured that maybe its test suite patched to buffer ops or some such)...

But I'm still seeing two failures (both in `test/put-get-del-test.js`):

```
# test put()/get()/del() with `null` value
ok 119 undefined
ok 120 no error, has key/value for ``null` value`
ok 121 is a Buffer
not ok 122 should be equal
  ---
    operator: equal
    expected: ''
    actual:   'null'
  ...
ok 123 no error, deleted key/value for ``null` value`
ok 124 entry propertly deleted
ok 125 should have correct error message
ok 126 value is undefined
# test put()/get()/del() with `undefined` value
ok 127 undefined
ok 128 no error, has key/value for ``undefined` value`
ok 129 is a Buffer
not ok 130 should be equal
  ---
    operator: equal
    expected: ''
    actual:   'undefined'
  ...
```

To get these tests to pass in pgdown I had to do this annoying `source == null ? '' : String(source)` thing, but I can't figure out where this was happening in the pre-2.5.0 code.

juliangruber:
> I think we should be more forceful on squashing PR's since it's difficult to know where something went wrong if some commits are a half finished feature or update.

+1

> Probably wouldn't be a bad idea to run the leveldown tests in the abstract-leveldown test suite to prevent this kind of thing -- if not in the test script itself, at least as a separate task in travis. Hell, it may not be a bad idea to run the levelup tests, and any other "reference" implementations.

tbh leveldown was using an undocumented api, so this kind of breakage is to be expected. What you proposed still would be awesome though, just to be sure.

PR #93, opened 2016-05-30
Run/fix tests in the browser

nolanlawson:
The abstract-leveldown tests currently fail in the browser. This fixes
the test and also adds new Travis tests using PhantomJS.

### DIFF
--- a/abstract/chained-batch-test.js
+++ b/abstract/chained-batch-test.js
@@ -185,10 +185,17 @@ module.exports.args = function (test) {
       if (Buffer.isBuffer(op.key)) op.key = String(op.key)
       if (Buffer.isBuffer(op.value)) op.value = String(op.value)
     })
-    t.deepEqual(ops, [
-        { type: 'put', key: '[object Object]', value: '[object Object]' }
-      , { type: 'del', key: '[object Object]' }
-    ])
+    var expected = process.browser ?
+      [
+          { type: 'put', key: '[object Object]', value: {beep: 'boop'} }
+        , { type: 'del', key: '[object Object]' }
+      ] :
+      [
+          { type: 'put', key: '[object Object]', value: '[object Object]' }
+        , { type: 'del', key: '[object Object]' }
+      ];
+    console.log('expected');
+    t.deepEqual(ops, expected)
     t.end()
   })
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,29 @@ language: node_js
 
 node_js:
   - "5"
-  - "4"
-  - "3"
-  - "2"
-  - "1.8"
-  - "1.0"
-  - "0.12"
-  - "0.10"
+
+script: npm run $COMMAND
+
+env:
+  matrix:
+    - COMMAND=test
+    - COMMAND=test-browser
+matrix:
+  include:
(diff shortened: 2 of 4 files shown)


nolanlawson:
I've see this error before; seem to recall it's related to Browserify. Can't remember what the fix is. :confused:

```
Error: Cannot find module 'node_modules/is-buffer/index.js' from '/home/travis/build/Level/abstract-leveldown'
    at /home/travis/build/Level/abstract-leveldown/node_modules/resolve/lib/async.js:46:17
    at process (/home/travis/build/Level/abstract-leveldown/node_modules/resolve/lib/async.js:173:43)
    at ondir (/home/travis/build/Level/abstract-leveldown/node_modules/resolve/lib/async.js:188:17)
    at load (/home/travis/build/Level/abstract-leveldown/node_modules/resolve/lib/async.js:69:43)
    at onex (/home/travis/build/Level/abstract-leveldown/node_modules/resolve/lib/async.js:92:31)
    at /home/travis/build/Level/abstract-leveldown/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:82:15)
```

vweevers:
While this PR might fix things for memdown, it could break another. Then it's an endless game of adding switches.

So how about only running these new `_serialize` tests on `abstract-leveldown` itself? Because it doesn't seem safe to assume, just on the basis of `process.browser`, what serialization the \*down adapter or underlying store performs (if any). Correct me if I'm wrong: to have that freedom is why `_serialize` was introduced, so it's weird if the tests "undo" that. /cc @juliangruber

abstract/chained-batch-test.js
```diff
@@ -185,10 +185,17 @@ module.exports.args = function (test) {
       if (Buffer.isBuffer(op.key)) op.key = String(op.key)
       if (Buffer.isBuffer(op.value)) op.value = String(op.value)
     })
-    t.deepEqual(ops, [
-        { type: 'put', key: '[object Object]', value: '[object Object]' }
-      , { type: 'del', key: '[object Object]' }
-    ])
+    var expected = process.browser ?
```
juliangruber: just stating for the record, we should never have `if (process.browser)` in this lib

juliangruber: am i right that this check is needed because we have `if (process.browser)` in lib code as well?

calvinmetcalf:
This https://github.com/defunctzombie/commonjs-assert/pull/22 is probably the fix

abstract/chained-batch-test.js
nolanlawson: Yeah, there are already [plenty of usages in here](https://github.com/Level/abstract-leveldown/search?utf8=%E2%9C%93&q=%22process.browser%22&type=Code).

nolanlawson:
@calvinmetcalf you mean a fix for the `is-buffer` issue?

ralphtheninja:
This PR sort of died. Not saying that we should merge it as is, but I would like to address these problems so we can get `memdown` and `level.js` tests to pass on latest `abstract-leveldown`. I have bandwidth and can spend time on fixing it, as long as we can come to some sort of agreement on what to do.

cc @juliangruber @nolanlawson @vweevers

PR #96, opened 2016-05-31
better null and undefined check for value serialization

stockulus:


nolanlawson:
This looks good to me, but I would be nervous not knowing which dependent projects it breaks and which ones it fixes. We know it fixes MemDOWN, but hard to know which other projects it breaks/fixes (localstorage-down, leveldown, etc.).

nolanlawson:
I suppose we could just merge it and then see which projects break. The alternative is to test each dependent, of which there are a lot according to npm.

dominictarr:
null and undefined may have valid encodings, for example in bytewise. That is already handled by levelup though right?

nolanlawson:
I haven't dug into why it's failing. In fact for memdown I just decided to peg it to abstract-leveldown 2.4.1 because that fixed the bug: https://github.com/Level/memdown/pull/51

nolanlawson:
I just tested manually, and this passes for leveldown and for memdown in Node, but fails for memdown in the browser. >_<

nolanlawson:
Looks like #85 is what broke memdown btw.

nolanlawson:
I don't have the energy to look into this right now. The tests are failing somehow, and somehow the serialization format is supposed to differ between browser and Node, but I can't figure out what the right values should be. I'm also nervous about inadvertently introducing breaking changes into PouchDB's adapters, because unfortunately we don't have tests in place to guard against inconsistencies in the on-disk storage format between localstorage-down/fruitdown versions (memdown is fine obviously).

If anybody else wants to test this out, just check out memdown, link it to this project, and then run `npm t` and `npm run test-browser-local`. I'm done; I'm just either going to pin to 2.4.1 permanently or fork abstract-leveldown.

nolanlawson:
Alternative fix: https://github.com/Level/memdown/pull/56

Issue #98, opened 2016-08-10
Put implementation converts `null` to `"null"` not `""`.


bigeasy:
There are new tests that assert that when a `null` or `undefined` value is put into the database that it will be retrieved as an empty string. I assume this means it should be converted to an empty string before insertion. The `_serializeValue` method converts using `String(value)` which converts `null` to `"null"` and `undefined` to `"undefined"`.

ralphtheninja:
See https://github.com/Level/abstract-leveldown/pull/119

PR #99, opened 2016-08-10
Convert nullish values to empty strings.

bigeasy:
Convert nullish values to empty strings `_serializeValue`. See #98.

### DIFF
--- a/test.js
+++ b/test.js
@@ -75,13 +75,17 @@ test('test key/value serialization', function (t) {
   t.equal(test._serializeKey(1), '1', '_serializeKey converts to string')
   t.ok(test._serializeKey(buffer) === buffer, '_serializeKey returns Buffer as is')
 
+  var browser = !! process.browser
+  process.browser = false
+
   t.equal(test._serializeValue(1), '1', '_serializeValue converts to string')
   t.equal(test._serializeValue(null), '', '_serializeValue converts null to empty string')
   t.ok(test._serializeValue(buffer) === buffer, '_serializeValue returns Buffer as is')
 
   process.browser = true
   t.equal(test._serializeValue(1), 1, '_serializeValue returns value as is when process.browser')
-  delete process.browser
+
+  process.browser = process.browser
 
   t.end()
 })
--- a/test.js
+++ b/test.js
@@ -62,6 +62,30 @@ test('test core extensibility', function (t) {
   ;new Test('foobar')
 })
 
+test('test key/value serialization', function (t) {
+  function Test (location) {
+    AbstractLevelDOWN.call(this, location)
+  }
+
+  util.inherits(Test, AbstractLevelDOWN)
+
+  var buffer = new Buffer(0)
+  var test = new Test('foobar')
+
+  t.ok(test._serializeKey(1) === '1', '_serializeKey converts to string')
+  t.ok(test._serializeKey(buffer) === buffer, '_serializeKey returns Buffer as is')
+
+  t.ok(test._serializeValue(1) === '1', '_serializeValue converts to string')
+  t.equal(test._serializeValue(null), '', '_serializeValue converts null to empty string')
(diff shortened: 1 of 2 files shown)


juliangruber:
LGTM, can you add a test though so we make sure this change stays?

bigeasy:
@juliangruber Actually, there is a test in the `abstract/put-get-del-test.js` file, but it is in the `nonErrorValues` function and the call to that function in `test.js` is commented out.

Should I add a collection of serialization tests in the "extensibility" section?

juliangruber:
hm we'd rather figure out why the test is commented out in the first place

bigeasy:
I would assume because it would require a concrete implementation to run those particular tests. None of the abstract tests that actually do anything against a concrete implementation are run in `test.js`. For example, the `open` test of `abstract/open-test.js` is not called. In fact, only the `args` test which tests the argument validation to the open function is called.

This is why I thought you might want to test just the serialization specifically, as a unit. It appears as though your counting on down streams to test an integration.

juliangruber:
ok yeah, that's very good reasoning, @bigeasy, I fully agree!

bigeasy:
`tape` documentation says that `t.same` is an alias for `t.deepEqual`. That's interesting. I'd expect it to be `===` strict quality. Didn't experiment.

Note that I set `process.browser = true` and then delete it. Let me know if I should preserve and reset. Do you actually run these tests in the browser or ever hope to?

juliangruber:
"equal" means it's the exactly equal thing, for example there can't be two equal bottle caps as each will be slightly different.

"same" means it's of the same kind, so for example two bottle caps of the same making are considered the same.

This is why `t.same` is an alias for `t.deepEqual`, as it's checking for same value and _not_ reference equality (=identity).

test.js
```diff
@@ -62,6 +62,30 @@ test('test core extensibility', function (t) {
...
+  function Test (location) {
+    AbstractLevelDOWN.call(this, location)
+  }
+
+  util.inherits(Test, AbstractLevelDOWN)
+
+  var buffer = new Buffer(0)
+  var test = new Test('foobar')
+
+  t.ok(test._serializeKey(1) === '1', '_serializeKey converts to string')
```
juliangruber: this should be written using `t.equal()`, so we get more useful error messages in case it doesn't match

test.js
```diff
@@ -62,6 +62,30 @@ test('test core extensibility', function (t) {
...
+  t.ok(test._serializeKey(buffer) === buffer, '_serializeKey returns Buffer as is')
+
+  t.ok(test._serializeValue(1) === '1', '_serializeValue converts to string')
+  t.equal(test._serializeValue(null), '', '_serializeValue converts null to empty string')
+  t.ok(test._serializeValue(buffer) === buffer, '_serializeValue returns Buffer as is')
+
+  process.browser = true
+  t.ok(test._serializeValue(1) === 1, '_serializeValue returns value as is when process.browser')
+  delete process.browser
```
juliangruber: this test suite might also be run inside the browser, so we have a problem here. i'd say the straight forward solution is to have `if (process.browser) { ... } else { ... }` inside the tests.

test.js
bigeasy: We're testing to see if converts to string, hence the strict equality, but I'll rewrite using `t.equal`.

bigeasy: Oh, now I see that `t.equal` is based on `===`.

bigeasy:
Getting closer, I think. Your thoughts @juliangruber?

juliangruber:
i think this looks good, but also i feel like i'm missing some background / prior discussion about null values. what do other maintainers think?

ralphtheninja:
@bigeasy @juliangruber I've rebased this onto latest master. Made a new PR https://github.com/Level/abstract-leveldown/pull/119

PR #100, opened 2016-09-19
fix put test on object serialization

juliangruber:
fixes https://github.com/Level/memdown/pull/57

Since abstract-leveldown implementers can choose to override `._serialize*` as they please, we really can't tell what kind of value we're getting. Simply checking for truthiness is like a smoke test for this behavior.

cc @nolanlawson @vweevers

### DIFF
--- a/abstract/put-test.js
+++ b/abstract/put-test.js
@@ -55,8 +55,8 @@ module.exports.args = function (test) {
     t.plan(3)
     var db = leveldown(testCommon.location())
     db._put = function (key, value, opts, callback) {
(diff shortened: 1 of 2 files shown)


vweevers:
Simple solution, +1.

Maybe we can also add an additional test [here](https://github.com/Level/abstract-leveldown/blob/master/test.js#L51), in line with the other "extensibility" tests? That would be a good place to test the default `_serialize` functions, as well as a custom `_serialize` - without tying it to any specific implementation.

juliangruber:
good call! i'll add those to this pr

nolanlawson:
:+1:

juliangruber:
@vweevers ^

PR #106, opened 2017-04-20
add checks for batch with null value

mvayngrib:
seems these both fail with leveldown, memdown and probably other implementations

what is the expected behavior? If there is no error on put, then it should probably mean it successfully wrote a null value.


vweevers:
Superseded by #277 (nullish values will be rejected).

PR #119, opened 2017-09-04
null and undefined serializes to empty string

ralphtheninja:
Rebased https://github.com/Level/abstract-leveldown/pull/99 and added test for `undefined`

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -254,9 +254,8 @@ AbstractLevelDOWN.prototype._serializeKey = function (key) {
 }
 
 AbstractLevelDOWN.prototype._serializeValue = function (value) {
-  return this._isBuffer(value) || process.browser || value == null
-    ? value
-    : String(value)
+  if (value == null) return ''
+  return this._isBuffer(value) || process.browser ? value : String(value)
 }
 
 AbstractLevelDOWN.prototype._checkKey = function (obj, type) {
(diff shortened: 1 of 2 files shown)


ralphtheninja:
This means `level.js` doesn't have to add extra code to make the `put-get-del-test.js` to pass

vweevers:
Why should null/undefined be serialized, rather than rejected?

ralphtheninja:
> Why should null/undefined be serialized, rather than rejected?

Because these tests https://github.com/Level/abstract-leveldown/blob/master/abstract/put-get-del-test.js#L140-L141

ralphtheninja:
`memdown` can also remove this code https://github.com/Level/memdown/blob/master/memdown.js#L148

vweevers:
I meant, taking a step back, why do we have this test - and behavior - in the first place? In streams `null` signifies end, so that might be a reason to reject storage of `null` values outright. But I know the history of all this is complicated and I can't immediately recall previous decisions.

ralphtheninja:
> I meant, taking a step back, why do we have this test - and behavior - in the first place?

_That_ is a very good question.

@juliangruber Hilfe.

I wouldn't mind dropping this and not allowing `null` nor `undefined`. But would mean a new major version. Right now I'm only concerned with fixing the damn tests ;)

PR #124, opened 2017-09-11
feat: add generic typings

MeirionHughes:
* serves as the basis types for levelUP's db typings. 
* while there is a lot of generics, they allow for levelup's typing to infer them directly from the db passed to it. 
* used locally with current levelup and passes levelup's tests. 
* I've added an indexer on each interface to derived classes can access anything not defined without typing error.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -35,7 +35,7 @@ declare namespace Abstract {
       TDeleteOptions=any,
       TIteratorOptions=any,
       TBatchOptions=any
-      >(db: any): LevelDOWN<
+      >(location: string): LevelDOWN<
       TKey,
       TValue,
       TOptions,
@@ -53,7 +53,7 @@ declare namespace Abstract {
       TDeleteOptions=any,
       TIteratorOptions=any,
       TBatchOptions=any
-      >(db: any): LevelDOWN<
+      >(location: string): LevelDOWN<
       TKey,
       TValue,
       TOptions,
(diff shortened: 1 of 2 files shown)


MeirionHughes:
Needs changes: `iterator` is missing from types.

abstract-leveldown.d.ts
```diff
@@ -14,5 +15,6 @@ export class AbstractLevelDOWN<
   put(key: TKey, value: TValue, options: TPutOptions, callback: any): void;
   del(key: TKey, options: TDeleteOptions, callback: any): void;
```
MeirionHughes: put and del `options` should be optional?

ralphtheninja: Yep!

ralphtheninja: @MeirionHughes Let me know when you are finished so I can release a new minor.

MeirionHughes:
Currently failing the levelup test.

MeirionHughes:
@ralphtheninja I think this is ready to go. bit of an odd behavour; where if you install `../abstract-leveldown` into `levelup`, the `levelup` tests fail. If I install normally and move these typescript declarations over manually, it all works.

abstract-chained-batch.d.ts
```diff
@@ -0,0 +1,12 @@
+export class AbstractChainedBatch<
```
MeirionHughes: these functions are typed wrong I think.

ralphtheninja:
2.7.0

PR #133, opened 2017-10-30
Fix implicit any errors in index.d.ts

### LINKED ISSUE #132: index.d.ts has lots of implicit any errors when compiled with --noImplicitAny
When using `leveldown` from typescript with `--noImplicitAny` (a subset of `--strict`), abstract-leveldown has 13 errors. This makes leveldown hard to consume for new projects written in Typescript.

sandersn:
Fixes #132


vweevers:
The failing test is unrelated to this PR. It fails on 0.12 because `sinon` [recently](https://github.com/sinonjs/sinon/pull/1583) added `supports-color` as a dependency, which is node 4+. I guess it's time to drop 0.12 here too. (cc @ralphtheninja).

vweevers:
3.0.0

PR #137, opened 2017-11-11
Fix: serialize batch

### LINKED ISSUE #136: Non-chained batch() doesn't serialize
Found this while working on serialization in `memdown`.

@juliangruber @ralphtheninja

vweevers:
Closes #136. Before this, the `_serializeKey` and `_serializeValue`  functions were only called if you used the chained batch form, not if you used the array form.

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -161,15 +161,14 @@ AbstractLevelDOWN.prototype.batch = function (array, options, callback) {
   if (!options || typeof options != 'object')
     options = {}
 
-  var i = 0
-    , l = array.length
-    , e
-    , err
+  var serialized = []
 
-  for (; i < l; i++) {
-    e = array[i]
-    if (typeof e != 'object')
+  for (var i = 0; i < array.length; i++) {
+    if (typeof array[i] !== 'object')
       continue
+
+    var e = Object.assign({}, array[i])
+    var err
 
     if (err = this._checkKey(e.type, 'type'))
       return callback(err)
@@ -177,16 +176,16 @@ AbstractLevelDOWN.prototype.batch = function (array, options, callback) {
     if (err = this._checkKey(e.key, 'key'))
       return callback(err)
 
-    // TODO: do not mutate input
     e.key = this._serializeKey(e.key)
 
-    if (e.type === 'put') {
+    if (e.type !== 'del')
       e.value = this._serializeValue(e.value)
-    }
+
+    serialized.push(e)
   }
 
   if (typeof this._batch == 'function')
-    return this._batch(array, options, callback)
+    return this._batch(serialized, options, callback)
 
   process.nextTick(callback)
 }
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -176,6 +176,13 @@ AbstractLevelDOWN.prototype.batch = function (array, options, callback) {
 
     if (err = this._checkKey(e.key, 'key'))
       return callback(err)
+
+    // TODO: do not mutate input
+    e.key = this._serializeKey(e.key)
+
+    if (e.type === 'put') {
+      e.value = this._serializeValue(e.value)
+    }
   }
 
   if (typeof this._batch == 'function')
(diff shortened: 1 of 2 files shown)


vweevers:
This breaks `memdown` in browsers, because [these](https://github.com/Level/abstract-leveldown/blob/24e11812d3e247bc2e6cb67955dd06614488333e/abstract/iterator-test.js#L10) abstract tests use numbers as values, `batch([..])` to write those values, and the behavior of `_serializeValue` is different if `process.browser` is true. 

That can be fixed quite easily, by not using numbers in those tests. I propose to establish a ground rule: abstract tests should exclusively use strings and buffers. Tests on any other types should be done in `encoding-down` or other stores if they support such types. 

WDYT @juliangruber @ralphtheninja?

Note as well that memdown currently has its own "serialization" in [its `_batch()` implementation](https://github.com/Level/memdown/blob/b51bbd30e1e16dd0d87b6acf40e863bb2eb37bc4/memdown.js#L197), with *again* a different behavior (no `process.browser` check here). Luckily, after we merge this PR and fix the abstract tests, we can remove the additional serialization from `memdown`, so that it all follows the same code path. We'll be far from done -- but at least the behavior of `put()` and `batch()` will be consistent. The next step would be to make browser and node consistent.

Perhaps another ground rule: `abstract-leveldown` is responsible for invoking serialization at the right time. The implementation should not (have to) do additional serialization in its `_put()` or `_batch()`.

vweevers:
To clarify *ground rules*. I think we need a clear set of guidelines to:

1. Help us fix all the weirdness (`null/undefined`, `process.browser`, empty strings and buffers, identity encoding, serialization of range options).
2. Have something to fall back on for future additions and feature requests.

juliangruber:
> That can be fixed quite easily, by not using numbers in those tests. I propose to establish a ground rule: abstract tests should exclusively use strings and buffers. Tests on any other types should be done in encoding-down or other stores if they support such types.

I agree 100%!

juliangruber:
Wow, you're killing it again!

abstract-leveldown.js
```diff
@@ -161,25 +161,31 @@ AbstractLevelDOWN.prototype.batch = function (array, options, callback) {
...
-  for (; i < l; i++) {
-    e = array[i]
-    if (typeof e != 'object')
+  var serialized = []
+
+  for (var i = 0; i < array.length; i++) {
+    if (typeof array[i] !== 'object')
       continue
 
+    var e = Object.assign({}, array[i])
```
vweevers: Is `Object.assign` supported in all our target environments, or should I use a ponyfill?

juliangruber: node 4 and up have it

vweevers: Browsers too, except IE.

juliangruber: edge has it, so imo we shouldn't worry

vweevers: IE is currently included in memdown's sauce labs matrix; should we remove it?

vweevers: It still has a [sizable market share](https://en.wikipedia.org/wiki/Usage_share_of_web_browsers#Summary_tables) esp. compared to Edge. I think we shouldn't be too quick to drop it (like we did with node 4).

vweevers:
We'll deal with IE later (in memdown to start, with docs and a polyfill in the test suite).

abstract-leveldown.js
```diff
@@ -161,25 +161,31 @@ AbstractLevelDOWN.prototype.batch = function (array, options, callback) {
...
-    , e
-    , err
-
-  for (; i < l; i++) {
-    e = array[i]
-    if (typeof e != 'object')
+  var serialized = []
+
+  for (var i = 0; i < array.length; i++) {
+    if (typeof array[i] !== 'object')
```
vweevers: Ooh, just realized this doesn't catch `null`

vweevers: #138

abstract-leveldown.js
vweevers: Level/memdown#102

PR #140, opened 2017-11-12
Abstract tests: use strings for values

vweevers:
As discussed in #137.

This means that memdown can do:

```js
MemDOWN.prototype._serializeKey = function (key) {
  return this._isBuffer(key)
    ? key
    : String(key)
}

MemDOWN.prototype._serializeValue = function (value) {
  if (value == null) return ''
  return value
}
```

Instead of the default implementation of `_serializeValue`:

```js
MemDOWN.prototype._serializeValue = function (value) {
  if (value == null) return ''
  return this._isBuffer(value) || process.browser ? value : String(value)
}
```

### DIFF
--- a/abstract/iterator-test.js
+++ b/abstract/iterator-test.js
@@ -8,13 +8,13 @@ var db
         d.push({
             type  : 'put'
           , key   : k
-          , value : Math.random()
+          , value : String(Math.random())
         })
       }
       return d
     }())
   , transformSource = function (d) {
-      return { key: d.key, value: String(d.value) }
+      return { key: d.key, value: d.value }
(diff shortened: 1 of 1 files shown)


vweevers:
I'm not saying (yet) that this is what `memdown` *should* do, we'll discuss that in `memdown`. But it could, is the point :)

vweevers:
Canary-tested against `leveldown` and my local branch of `memdown` in node and chrome. It fails with TypeScript but that seems to be an issue with `npm link` (Level/memdown#101).

vweevers:
Release as patch?

PR #142, opened 2017-11-25
Batch array: assert elements are objects

### LINKED ISSUE #138: Remove gaps from batch array?
If you do `db.batch([null])`, or any other falsy value, should `abstract-leveldown` filter it?

I noticed that memdown has its own `!array[i]` check in `_batch()`, so I wondered if that's a job for `abstract-leveldown`. We currently do have a `typeof array[i] !== 'object'` check, which I didn't git blame yet, but this doesn't catch `null`.

@juliangruber @ralphtheninja

vweevers:
Closes #138.

Breaking change, major.

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -161,11 +161,11 @@ AbstractLevelDOWN.prototype.batch = function (array, options, callback) {
   if (!options || typeof options != 'object')
     options = {}
 
-  var serialized = []
+  var serialized = new Array(array.length)
 
   for (var i = 0; i < array.length; i++) {
-    if (typeof array[i] !== 'object')
-      continue
+    if (typeof array[i] !== 'object' || array[i] === null)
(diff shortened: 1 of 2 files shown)


juliangruber:
For reference, pre-allocating the array is the fastest method [for big](https://jsperf.com/grow-array) and [small arrays](https://jsperf.com/grow-array-small/)

ralphtheninja:
I just noticed that we're calling back on the same tick on multiple places. Creating new issue for this.

vweevers:
@ralphtheninja also, my first instinct was to throw (because it's a programmer's error), instead of calling back with an error. But then I thought it should be the same as the lines below:

https://github.com/Level/abstract-leveldown/blob/a5069de9deef13fd1010e5d7b67107211cd48235/abstract-leveldown.js#L173-L174

What's our policy on that?

vweevers:
Never mind, I realized we do that in order to wrap the error in levelup.

PR #144, opened 2017-11-25
Remove test: serialize object key in approximateSize()

### LINKED ISSUE #139: Remove test: serialization of object key in approximateSize()
I propose to remove this abstract test:

https://github.com/Level/abstract-leveldown/blob/20093fba5d194b857474be3a4128a6a800268eef/abstract/approximate-size-test.js#L70-L81

We have enough coverage on the `_serialize` functions, including the extensibility of `_serializeKey()` in combination with `approximateSize()`. This test has no added value anymore and assumes that object keys are stringified.

vweevers:
Closes #139.

### DIFF
--- a/abstract/approximate-size-test.js
+++ b/abstract/approximate-size-test.js
@@ -67,19 +67,6 @@ module.exports.args = function (test) {
     t.end()
   })
 
-  test('test _serialize object', function (t) {
-    t.plan(3)
-    var db = leveldown(testCommon.location())
-    db._approximateSize = function (start, end, callback) {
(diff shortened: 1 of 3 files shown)


vweevers:
Also remove:

- [ ] `del`: https://github.com/Level/abstract-leveldown/blob/e73dee7b82dd0f392edc399d29419b47bc491edc/abstract/del-test.js#L45
- [ ] `get`: https://github.com/Level/abstract-leveldown/blob/e73dee7b82dd0f392edc399d29419b47bc491edc/abstract/get-test.js#L45

Issue #146, opened 2017-11-26
Move batch type defaulting from levelup to here?


vweevers:
By type defaulting, if that's even a word, I mean  assigning a default type of `put` to batch operations.

If we move [this levelup logic](https://github.com/Level/levelup/blob/8470b1e7328aff560e020f82437b2af91dc43a5f/lib/levelup.js#L259-L262) to `abstract-leveldown`, we don't need to iterate the batch array twice.

WDYT @juliangruber @ralphtheninja?

juliangruber:
That's clever, works for me!

vweevers:
Hmm, the current logic:

```js
if (!op.type && op.key !== undefined && op.value !== undefined) { op.type = 'put' }
```

Implies we don't support `undefined` values, though in other places we do. An operation like this won't get a type:

```js
db.batch([ { key: 'a', value: undefined }]
```

We have 5 options:

1. Check for property existence: `if (!op.type && 'key'  in op && 'value'  in op) op.type = 'put'`
2. Always default to put: `if (!op.type) op.type = 'put'`
3. No default: `if (!op.type) throw new Error(..)`
4. Even more strict: `if (op.type !== 'put' && op.type !== 'del') throw new Error(..)`
5. Definitively say we don't support `undefined` values (and not just here).

I think I prefer throwing (3 or 4), avoids surprises and cleverness.

ralphtheninja:
I prefer 4. The code makes it clear what type of operations a batch operation supports.

vweevers:
@ralphtheninja IIRC this is all done and released now. Let's close?

ralphtheninja:
:boom:

PR #148, opened 2017-12-02
Batch: assert type is 'put' or 'del'

vweevers:
See #146.

Sometime later, we should remove levelup's logic for this.


vweevers:
I think this is the last thing before a new major.

Issue #115, opened 2017-08-13
add standard


ralphtheninja:
Don't forget to add standard badge.

Issue #149, opened 2017-12-04
Remove _isBuffer?


vweevers:
Implementations no longer have a need to override `_isBuffer` in order to use typed arrays as buffers. Because with modern browserify, Buffer is a subclass of Uint8Array and conversions between the two are cheap. I.e. we can use Buffer (and `Buffer.isBuffer`) everywhere.

@ralphtheninja @juliangruber

vweevers:
Tasks:

- [ ] `abstract-leveldown`: replace usage of `_isBuffer` with `Buffer.isBuffer`
- [ ] `abstract-leveldown`: remove `AbstractLevelDOWN.prototype._isBuffer`
- [ ] new release
- [ ] do the same for implementations (https://github.com/search?q=org%3ALevel+_isBuffer&type=Code)

vweevers:
Maybe we can also remove `testBuffer` used in `put-get-del-test`?

https://github.com/Level/abstract-leveldown/blob/49a1bd7156f3c476eda8f67e0f79946ce3674f3e/abstract/put-get-del-test.js#L171-L172

I don't know the history of that one but assume it was also to enable use of replacement Buffer types.

@ralphtheninja

ralphtheninja:
@vweevers Aye, I created an issue for it here https://github.com/Level/abstract-leveldown/issues/126 . Will look into it next.

ralphtheninja:
@vweevers Should we remove `approximateSize` for version 4 as well?

vweevers:
> Should we remove approximateSize for version 4 as well?

Sure why not :) Which other issues do you want to attack though? Note I didn't assign the ones I opened to the milestone you created.

ralphtheninja:
@vweevers It depends on if we have more breaking changes coming. Adding tests and tweaking we should be able to release as minors. I just want to avoid ending up in a position where we implement for ages and never release :)

ralphtheninja:
That said, it might also be worth it to spend time on releasing v4, since implementations rarely update `abstract-leveldown`. Trade offs :)

vweevers:
Both true. I want to settle somewhere in between, in order to continue with `memdown`, my fork of `localstorage-down` and eventually `level-js`.

vweevers:
@ralphtheninja I pushed my work to [webstorage-down](https://github.com/vweevers/webstorage-down) if you're interested.

ralphtheninja:
[...]

PR #150, opened 2017-12-17
add standard

ralphtheninja:
Resolves https://github.com/Level/abstract-leveldown/issues/115

Many tiny commits. Will squash later.


abstract/chained-batch-test.js
```diff
@@ -166,7 +166,7 @@ module.exports.args = function (test) {
     var batch = db.batch().put('foo', 'bar')
     batch.write(function () {})
     try {
-      batch.write(function (err) {})
```
ralphtheninja: Fixes the `Expected error to be handled`.

abstract/put-get-del-test.js
```diff
@@ -110,7 +110,7 @@ module.exports.errorKeys = function (testFunc, BufferType) {
   makeErrorKeyTest('null key', null, /key cannot be `null` or `undefined`/)
   makeErrorKeyTest('undefined key', undefined, /key cannot be `null` or `undefined`/)
   makeErrorKeyTest('empty String key', '', /key cannot be an empty String/)
-  makeErrorKeyTest('empty Buffer key', new BufferType(0), /key cannot be an empty \w*Buffer/)
+  makeErrorKeyTest('empty Buffer key', BufferType.alloc(0), /key cannot be an empty \w*Buffer/)
```
ralphtheninja: This might be an issue since tests can pass in a custom `BufferType`, which defaults to `Buffer` if not set.

abstract/put-get-del-test.js
```diff
@@ -157,7 +157,7 @@ module.exports.nonErrorValues = function (testFunc, BufferType) {
   makePutGetDelSuccessfulTest('`null` value', 'foo null', null, '')
   makePutGetDelSuccessfulTest('`undefined` value', 'foo undefined', undefined, '')
   makePutGetDelSuccessfulTest('empty String value', 'foo', '', '')
-  makePutGetDelSuccessfulTest('empty Buffer value', 'foo', new BufferType(0), '')
+  makePutGetDelSuccessfulTest('empty Buffer value', 'foo', BufferType.alloc(0), '')
```
ralphtheninja: As above.

ralphtheninja:
I have tested with `leveldown`. Maybe we should do `npm link abstract-leveldown` on a few implementations to see if there's any problems? E.g. `memdown` etc.

abstract/chained-batch-test.js
vweevers: Which issue are you referring to?

abstract-leveldown.js
```diff
@@ -29,7 +29,7 @@ AbstractLevelDOWN.prototype.open = function (options, callback) {
 
   if (typeof options !== 'object') { options = {} }
 
-  options.createIfMissing = options.createIfMissing != false
+  options.createIfMissing = options.createIfMissing !== false
```
vweevers: This is a breaking change, to go from "not falsy" to "not false". No strong feelings here, but let's make it a conscious group decision.

test.js
```diff
@@ -53,13 +53,13 @@ require('./abstract/iterator-test').sequence(test)
...
     AbstractLevelDOWN.call(this, location)
-    t.equal(this.location, location, 'location set on `this`')
-    t.end()
   }
 
   util.inherits(Test, AbstractLevelDOWN)
 
-  ;new Test('foobar')
+  var test = new Test('foobar')
+  t.equal(test.location, 'foobar', 'location set on `this`')
```
vweevers: Maybe update the assertion message to `location set on instance` (same below)

abstract/put-get-del-test.js
vweevers: There probably isn't a need for that anymore, same as #149

vweevers:
> Maybe we should do npm link abstract-leveldown on a few implementations to see if there's any problems? E.g. memdown etc.

I'll test against memdown (v2) in a bit

vweevers:
Memdown (node+browsers) is green: https://travis-ci.org/Level/memdown/builds/317637331.

@ralphtheninja if you make changes, you can simply rerun the above build.

testCommon.js
```diff
@@ -28,6 +28,7 @@ var cleanup = function (callback) {
 
     list.forEach(function (f) {
       rimraf(path.join(__dirname, f), function (err) {
+        if (err) return callback(err)
```
vweevers: ⚠️ This breaks leveldown+rocksdb on Windows, which may or may not be a good thing

vweevers: I suspected this is because the tests do not close the db. Confirmed with a hack in `abstract-leveldown` that closes the last known db before calling `rimraf`:

```js
var cleanup = function (callback) {
  // ..

  if (global.__db && global.__db.status === 'open') {
    global.__db.close(function (err) {
      if (err) throw err
      cleanup(callback)
    })

    global.__db = null
    return
  }

  // ..
}
```

With this, the leveldown tests are passing on Windows.

abstract/chained-batch-test.js
ralphtheninja: Oh, there's no issue. I meant standard complaining about error not being handled.

testCommon.js
ralphtheninja: Ok, so we might as well skip the `err` parameter completely and do nothing.

vweevers: Sure. Maybe open an issue to tackle closing the db's later

abstract/put-get-del-test.js
ralphtheninja: Do you mean that the test is no longer needed?

vweevers: I mean we can probably remove the `BufferType` argument in `exports.errorKeys(testFunc, BufferType)`. And just do `Buffer.alloc(0)`.

vweevers: Added it to the tasklist in #149.

abstract-leveldown.js
ralphtheninja: True. I made a quick gist

```js
function foo (opts) {                        
  console.log(opts.createIfMissing != false) 
  console.log(opts.createIfMissing !== false)
}                                            
                                             
foo({})                                      
foo({ createIfMissing: null })               
foo({ createIfMissing: '' })                 
```

Gives:

```
true
true
true
true
false
true
```

ralphtheninja: I personally prefer it to be "not false".

vweevers: I'm on board

vweevers:
Rerunning appveyor builds.

abstract-chained-batch.js
```diff
@@ -15,24 +15,23 @@ AbstractChainedBatch.prototype._serializeValue = function (value) {
 }
 
 AbstractChainedBatch.prototype._checkWritten = function () {
-  if (this._written)
-    throw new Error('write() already called on this batch')
+  if (this._written) { throw new Error('write() already called on this batch') }
```
vweevers: I personally prefer multi-line if-statements for long lines like these. Does `standard` not enforce this because of opinion, or because of technical limitations? Just curious. As you know I tried `prettier-standard` on `memdown` but its output wasn't much better.

abstract-leveldown.js
```diff
@@ -219,28 +216,27 @@ AbstractLevelDOWN.prototype._setupIteratorOptions = function (options) {
   options = xtend(options)
 
   ;[ 'start', 'end', 'gt', 'gte', 'lt', 'lte' ].forEach(function (o) {
-    if (options[o] && self._isBuffer(options[o]) && options[o].length === 0)
+    if (options[o] && self._isBuffer(options[o]) && options[o].length === 0) {
```
vweevers: Side notes:

1. `options[o] && self._isBuffer(options[o])` can be shortened to `self._isBuffer(options[o])`
2. Shouldn't it also remove empty strings and `null`? Because these should all behave the same way, at least according to the current tests. If we remove those options here, implementations don't need to worry about it.

abstract/ranges-test.js
```diff
@@ -13,35 +13,35 @@ module.exports.setUp = function (leveldown, test, testCommon) {
 module.exports.iterator = function (leveldown, test, testCommon, collectEntries) {
   test('test simple iterator()', function (t) {
```
vweevers: Side note 300: we have duplicate tests in `iterator-test.js` and `ranges-test.js`.

abstract-chained-batch.js
ralphtheninja: I don't know why `standard --fix` puts them on one line.

I do prefer multi-line if-statements as well.

abstract/ranges-test.js
ralphtheninja: Make an issue for it so we can clean it up.

abstract-leveldown.js
ralphtheninja: Well, you obviously already made some notes of it ;)

abstract-leveldown.js
ralphtheninja: @juliangruber Any opinion on this one?

abstract-leveldown.js
vweevers: You weren't kidding about side quests ;)

ralphtheninja:
@vweevers Btw, really nice idea to just trigger builds for dependent modules. Is there any tool for this? Or did you do it manually?

vweevers:
@ralphtheninja manually. It was an experiment to do it in some structured way (while thinking: could a tool do this)

ralphtheninja:
I was thinking exactly that :) Sort of what Greenkeeper does, but point out github branches rather than version numbers on npm.

testCommon.js
ralphtheninja: I'm thinking we could put back `if (err) return callback(err)` before working on closing dbs, once the tests pass on Windows we know we should be good.

Will do this on a separate branch and have a temporary branch in `leveldown` to track the build status (`abstract-leveldown-v4` branch).

Issue #126, opened 2017-09-12
Clean up test and testBuffer globals


ralphtheninja:
~~This is an investigative issue, will add points as I find them.~~

Get rid of `test` and `testBuffer` "globals", pass them on as function parameters instead

Try to get rid of `testBuffer` completely, if possible.

ralphtheninja:
Fix https://github.com/Level/abstract-leveldown/issues/155 first.

vweevers:
Ah I missed this. 👍

ralphtheninja:
Ok, here's some code archeology.

I cloned the repos from https://github.com/level/awesome#stores

![downs](https://user-images.githubusercontent.com/308049/34084900-74e5bb44-e388-11e7-9a8e-efecfff20e24.png)

A search for `testBuffer` gave:

![testbuffer](https://user-images.githubusercontent.com/308049/34084957-4d6796e0-e389-11e7-90c1-de23e857fece.png)

It seems to me that all of them are passing in a plain `Buffer` object, although the content is different. Some pass in `new Buffer('somestring')` and some pass in data read from file.

The `testBuffer` variable is then used in two different tests:

* `.nonErrorKeys` where `testBuffer` is used as a _key_
* `.nonErrorValues` where `testBuffer` is used as a _value_

`testBuffer` doesn't seem to have anything to do with `BufferType`.

Conclusion:

Unless the actual content of `testBuffer` is important (doesn't seem likely) we could swap out `testBuffer` variable to a constant inside `abstract-leveldown`.

@vweevers Let me know what you think.

@juliangruber Any thoughts?

ralphtheninja:
A few side notes:

* why is it important for implementations to be able to pass in `testFunc`? why not just use `tape`?
* ditto for the `testCommon` parameter, why not use `testCommon.js` inside `abstract-leveldown`?

ralphtheninja:
Ideally the tests used in implementations should be (if possible)

```js
require('abstract-leveldown/abstract/put-get-del-test').all(whateverdown)
```

vweevers:
> why is it important for implementations to be able to pass in testFunc? why not just use tape?

Because if they have additional custom tests, they must use the same `tape`. Otherwise, there'd be 2 test harnesses, executing tests at the same time.

PR #159, opened 2017-12-18
test and testBuffer as parameters

ralphtheninja:
Resolves https://github.com/Level/abstract-leveldown/issues/126

Note that this doesn't remove `testBuffer`.

~~`db` is also a local global that could be called back in e.g. `.setUp(leveldown, test, testCommon, cb)` and later passed on as parameter to underlying functions.~~

### DIFF
--- a/abstract/put-get-del-test.js
+++ b/abstract/put-get-del-test.js
@@ -1,5 +1,4 @@
 var db
-var testBuffer
 var verifyNotFoundError = require('./util').verifyNotFoundError
 
 function makeGetDelErrorTests (test, type, key, expectedError) {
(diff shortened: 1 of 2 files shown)


ralphtheninja:
@vweevers WDYT? Should I pass `db` as a parameter as well while I'm at it?

vweevers:
> WDYT? Should I pass db as a parameter as well while I'm at it?

Not sure. How would it work if an implementation doesn't use `all`, but this pattern:

```js
require('abstract-leveldown/abstract/put-get-del-test').setUp(leveldown, test, testCommon)
require('abstract-leveldown/abstract/put-get-del-test').errorKeys(test)
require('abstract-leveldown/abstract/put-get-del-test').nonErrorKeys(test, testBuffer)
require('abstract-leveldown/abstract/put-get-del-test').errorValues(test)
require('abstract-leveldown/abstract/put-get-del-test').nonErrorValues(test, testBuffer)
require('abstract-leveldown/abstract/put-get-del-test').tearDown(test, testCommon)
```

ralphtheninja:
> Not sure. How would it work if an implementation doesn't use all, but this pattern:

It wouldn't I guess, unless `.setUp`.

ralphtheninja:
Hmm, what about something like this?

```js
const suite = require('abstract-leveldown/abstract/put-get-del-test')
const test = suite.setUp(leveldown, tape, testCommon, testBuffer)
test.errorKeys()
test.nonErrorKeys()
test.errorValues()
test.nonErrorValues()
test.tearDown()
```

ralphtheninja:
@vweevers Ok to merge?

Issue #152, opened 2017-12-17
Cleanup duplicate tests


vweevers:
In `iterator-test.js` and `ranges-test.js`. See https://github.com/Level/abstract-leveldown/pull/150#discussion_r157370377

ralphtheninja:
So it seems the file `ranges-test.js` is completely obsolete since it only has one main test (`module.exports.iterator`) apart from `.setUp` and `.tearDown`. So will remove the complete file.

This makes me wonder a bit why this file was added at all in the first place.

ralphtheninja:
I made a diff of `module.exports.iterator()` methods from `iterator-test.js` and `ranges-test.js` and they aren't identical.

Diff below corresponds to output of `diff iterator-test.js ranges-test.js` but only from `module.exports.iterator`.

`ranges-test.js` is more modern since it tests `lt`, `gt` etc ranges and not the legacy `start`, `end`

```diff
16,17d15
<           t.ok(Buffer.isBuffer(key), 'key argument is a Buffer')
<           t.ok(Buffer.isBuffer(value), 'value argument is a Buffer')
73,74c71,72
<   test('test iterator with start=0', function (t) {
<     collectEntries(db.iterator({ keyAsBuffer: false, valueAsBuffer: false, start: '00' }), function (err, data) {
---
>   test('test iterator with gte=0', function (t) {
>     collectEntries(db.iterator({ keyAsBuffer: false, valueAsBuffer: false, gte: '00' }), function (err, data) {
83,84c81,82
<   test('test iterator with start=50', function (t) {
<     collectEntries(db.iterator({ keyAsBuffer: false, valueAsBuffer: false, start: '50' }), function (err, data) {
---
>   test('test iterator with gte=50', function (t) {
>     collectEntries(db.iterator({ keyAsBuffer: false, valueAsBuffer: false, gte: '50' }), function (err, data) {
93,94c91,92
<   test('test iterator with start=50 and reverse=true', function (t) {
<     collectEntries(db.iterator({ keyAsBuffer: false, valueAsBuffer: false, start: '50', reverse: true }), function (err, data) {
---
>   test('test iterator with lte=50 and reverse=true', function (t) {
>     collectEntries(db.iterator({ keyAsBuffer: false, valueAsBuffer: false, lte: '50', reverse: true }), function (err, data) {
104c102
<     collectEntries(db.iterator({ keyAsBuffer: false, valueAsBuffer: false, start: '49.5' }), function (err, data) {
---
[...]

Issue #157, opened 2017-12-17
Export simple interface to entire abstract suite


vweevers:
I'm working on a `localstorage-down` alternative (not because I need it, but was halfway done for PouchDB, and figured I see it through to get a new perspective on the whole buffer/browser/serialize story).

It cannot offer snapshot guarantees, which is fine if documented. But if you need to skip tests, like snapshots, you need to replace:

```js
require('abstract-leveldown/abstract/iterator-test').all(leveldown, test, testCommon)
```

With 5 lines more:

```js
require('abstract-leveldown/abstract/iterator-test').setUp(leveldown, test, testCommon);
require('abstract-leveldown/abstract/iterator-test').args(test);
require('abstract-leveldown/abstract/iterator-test').sequence(test);
require('abstract-leveldown/abstract/iterator-test').iterator(leveldown, test, testCommon, testCommon.collectEntries);
// require('abstract-leveldown/abstract/iterator-test').snapshot(leveldown, test, testCommon);
require('abstract-leveldown/abstract/iterator-test').tearDown(test, testCommon);
```

Or be clever about it:

```js
require('abstract-leveldown/abstract/iterator-test').snapshot = function() {}
require('abstract-leveldown/abstract/iterator-test').all(leveldown, test, testCommon)
```

Neither is ideal. Same story with other tests. In addition, some tests were missing from `localstorage-down`. No one would know.

So I was thinking: can we export a simple function from `abstract/index.js`, that runs all tests by default, takes the usual arguments, as well as options to exclude tests? Something like:

```js
require('abstract-leveldown/abstract')(leveldown, test, testCommon, {
  'open': {
    skip: 'openAdvanced'
  }
  'iterator': {
    skip: 'snapshot'
  },
  'approximate-size': {
    skip: 'all'
  }
})
```

Or move frequently skipped tests to separate files so we don't need nested options:

```js
require('abstract-leveldown/abstract')(leveldown, test, testCommon, {
  skip: [
    'open-advanced',
    'snapshot',
    'approximate-size'
  ]
})
```

Also makes it really easy to see what an implementation *doesn't* support.

@ralphtheninja @juliangruber

ralphtheninja:
I like the idea, but would like to simplify (or improve) `test` and `testCommon` as well. E.g. something like (building on top of your idea):

```js
require('abstract-leveldown/abstract')(leveldown, {
  test: require('tap'), // optional, defaults to tape
[...]

Issue #151, opened 2017-12-17
Make sure dbs are closed properly in tests


ralphtheninja:
See:

- https://github.com/Level/abstract-leveldown/pull/150#discussion_r157366010 (not closing db's results in rimraf `EBUSY` errors on Windows)
- https://github.com/Level/abstract-leveldown/pull/161#issuecomment-353202605 (missing `teardown` calls).
- https://github.com/Level/abstract-leveldown/pull/161#issuecomment-353202605 (missing close)

PR #161, opened 2017-12-20
Cleanup iterator tests

ralphtheninja:
Resolves https://github.com/Level/abstract-leveldown/issues/152


abstract/iterator-test.js
```diff
@@ -178,7 +178,6 @@ module.exports.iterator = function (leveldown, test, testCommon, collectEntries)
   test('test full data collection', function (t) {
     collectEntries(db.iterator({ keyAsBuffer: false, valueAsBuffer: false }), function (err, data) {
       t.error(err)
-      t.equal(data.length, sourceData.length, 'correct number of entries')
```
vweevers: Can be useful info for an implementer though, if the data isn't equal, and the problem is in the length. Most likely it will be, for these range tests.

(btw, very nice work so far)

ralphtheninja: Check. Will put it back once done with refactoring.

abstract/iterator-test.js
```diff
@@ -182,6 +182,7 @@ module.exports.iterator = function (leveldown, test, testCommon, collectEntries)
     test(name, function (t) {
       collectEntries(db.iterator(opts), function (err, data) {
         t.error(err)
+        t.is(data.length, expected.length, 'correct number of entries')
         t.same(data, expected)
```
ralphtheninja: @vweevers Here it is again :grinning:

abstract/iterator-test.js
```diff
@@ -189,457 +189,269 @@ module.exports.iterator = function (leveldown, test, testCommon, collectEntries)
...
+    lt: '9b',
+    gt: '9a',
+    reverse: true
+  }, [])
+
+  doTest('test iterator with start and end after database and reverse=true - legacy', {
+    start: '9b',
+    end: '9a',
+    reverse: true
+  }, [])
```
ralphtheninja: This diff might be difficult to read since many lines have disappeared.

vweevers: I'll review commit by commit. So far I applaud every change you've made.

Looking at the end result, and the tests we're left with, I wonder if the whole thing would be more focused if you do put it in two files after all. I.e. move all the range tests (those using `doTest`) to a separate file - but instead of `ranges-test.js` maybe name it `iterator-range-test.js`.

Then we'd have:

1. `iterator-test.js` doing basic API tests
2. `iterator-range-test.js` doing actual iteration tests (exclusively through `doTest`).

Both would have their own setup and teardown, and run on an isolated db.

vweevers: (maybe that was the original intention for these two files?)

ralphtheninja: Good idea. So `iterator-range-test.js` could have its own `module.exports.setUp` that does what the `'setUp #2'` does (we will also get rid of the re-open code before batching the test data). Also, maybe `doTest` should be called `rangeTest` instead.

vweevers: Separating these tests into files (by concern / feature) also means it's easier to skip them when needed. Perhaps then we don't need #157. Or if we do, then that wrapper will be simpler as well; it'd only need to skip files.

ralphtheninja: Updated. Check it out.

abstract/iterator-range-test.js
```diff
@@ -0,0 +1,375 @@
...
+  var collectEntries = testCommon.collectEntries
+
+  function rangeTest (name, opts, expected) {
+    opts.keyAsBuffer = false
+    opts.valueAsBuffer = false
+    test(name, function (t) {
+      collectEntries(db.iterator(opts), function (err, data) {
+        t.error(err)
+        t.is(data.length, expected.length, 'correct number of entries')
+        t.same(data, expected)
```
ralphtheninja: Hmm, maybe this local `data` is confusing with outer `data`.

ralphtheninja: Maybe call it `result` instead?

ralphtheninja:
Maybe even split into two range test methods, `.range` and `.rangeLegacy`? Or are we starting to split hairs now? :)

vweevers:
> Maybe even split into two range test methods, .range and .rangeLegacy?

I think we can do that if and when we remove the legacy ranges. I.e. when we have two camps: implementations that still support the legacy ranges, and those that don't.

vweevers:
Is this missing `tearDown`?:

https://github.com/Level/abstract-leveldown/blob/49054256d67dd98f2fd76b1844723ea0ff244ffe/test.js#L48-L51

ralphtheninja:
> I think we can do that if and when we remove the legacy ranges. I.e. when we have two camps: implementations that still support the legacy ranges, and those that don't.

I took a quick look at the `start`, `end`, `lt`, `lte` code in `leveldown` and it's quite hairy to say the least.

vweevers:
Yea let's not touch that right now ;)

ralphtheninja:
> Is this missing tearDown?:

It looks like there are multiple missing `.tearDown`

ralphtheninja:
Maybe create a separate issue for this. I noticed that the `db` isn't closed properly here:

https://github.com/Level/abstract-leveldown/blob/6d9ac3244792ce33c5e1affaf57b76ba8fba11c6/abstract/del-test.js#L52-L56

vweevers:
I'll add it to #151.

ralphtheninja:
Good to squash and merge?

vweevers:
> Good to squash and merge?

I need a little bit more time. Maybe smoke test it against leveldown in the meantime?

ralphtheninja:
> I need a little bit more time. Maybe smoke test it against leveldown in the meantime?

Take your time. I have tested it against leveldown, but breaking changes since `ranges-test.js` is gone.

PR #168, opened 2017-12-23
Buffer.isBuffer()

ralphtheninja:
Resolves https://github.com/Level/abstract-leveldown/issues/149


vweevers:
@ralphtheninja this change results in a `Cannot find module '../../is-buffer/index.js'` error when browserified. Don't know yet if it's a browserify or zuul problem, will update.

vweevers:
Most likely zuul, see also https://github.com/browserify/browserify/issues/1483#issuecomment-197525841

vweevers:
Looks like it only happens when `abstract-leveldown` is npm-linked. Given the complexity of this issue, I'll avoid npm link for now and *maybe* dive into it later.

PR #171, opened 2017-12-23
Remove object value test

### LINKED ISSUE #170: Remove object value test
For the same reasons as #158 and #144.

https://github.com/Level/abstract-leveldown/blob/d3a4022aae9bc2ec004c1b213c48ba55f3963819/abstract/put-test.js#L100-L111

vweevers:
Closes #170.

### DIFF
--- a/abstract/put-test.js
+++ b/abstract/put-test.js
@@ -96,19 +96,6 @@ module.exports.put = function (test) {
       })
     })
   })
-
-  if (process.browser) {
-    test('test object value put()', function (t) {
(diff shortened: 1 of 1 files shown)


abstract/put-test.js
```diff
@@ -96,19 +96,6 @@ module.exports.put = function (test) {
       })
     })
   })
-
-  if (process.browser) {
```
ralphtheninja: Lovely to get rid of a `process.browser` check!

Issue #154, opened 2017-12-17
_setupIteratorOptions should also delete empty strings and null


vweevers:
In addition to deleting empty Buffers:

https://github.com/Level/abstract-leveldown/blob/6edb4f8659e389e073ffb0a8d8c2dde694ec827a/abstract-leveldown.js#L216-L224

Because an empty Buffer, empty string and `null` should all behave the same way, at least according to the current tests. If we remove those options here, implementations don't need to worry about it.

ralphtheninja:
I really don't like the `delete` thingie at all. I would prefer if this functionality could be lifted out into a function (it's not immediately clear to me what it does) which doesn't mutate the parameter. It should return a new object instead.

While writing it I realized `xtend()` here ensures that we aren't mutating the original `options` parameter so we're good, but it should be part of the function that xtends/mutates the original `options`.

ralphtheninja:
Nevermind, what I just said above is exactly what the `._setupIteratorOptions` does already. :grinning:

vweevers:
@ralphtheninja I think you're right though about:

> it should be part of the function that xtends/mutates the original options.

We could do:

```js
var hasOwnProperty = Object.prototype.hasOwnProperty;

AbstractLevelDOWN.prototype._setupIteratorOptions = function (options) { 
  var clean = {}

  for (var k in options) {
    if (!hasOwnProperty.call(options, k)) continue
    if (isRangeOption(k) && isEmptyRangeOption(options[k])) continue
    
    clean[k] = options[k]
  }

  return clean
}

function isRangeOption (k) {
  return (
    k === 'gt' || k === 'gte' ||
    k === 'lt' || k === 'lte' ||
    k === 'start' || k === 'end'
  )
}

function isEmptyRangeOption (v) {
  return v === '' || v == null || isEmptyBuffer(v)
}

function isEmptyBuffer (v) {
  return Buffer.isBuffer(v) && v.length === 0
}
```

ralphtheninja:
> We could do:

I like it! Will update the PR.

ralphtheninja:
Just realized I never made the PR, but my code pretty much looks like yours. Will push it up after https://github.com/Level/abstract-leveldown/pull/175

PR #176, opened 2018-01-06
Setup iterator options

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/154

Will squash this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,17 @@ By default a `batch()` operation without arguments returns a blank `AbstractChai
 
 By default an `iterator()` operation returns a blank `AbstractIterator` object. The prototype is available on the main exports for you to extend. If you want to implement iterator operations then you should extend the `AbstractIterator` and return your object in the `_iterator(options)` method.
 
+The `iterator()` operation accepts the following range options:
+
+* `gt`
+* `gte`
+* `lt`
+* `lte`
+* `start` (legacy)
+* `end` (legacy)
+
+A range option that is either an empty buffer, an empty string or `null` will be ignored.
+
 `AbstractIterator` implements the basic state management found in LevelDOWN. It keeps track of when a `next()` is in progress and when an `end()` has been called so it doesn't allow concurrent `next()` calls, it does allow `end()` while a `next()` is in progress and it doesn't allow either `next()` or `end()` after `end()` has been called.
 
 ### AbstractIterator(db)
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -1,8 +1,10 @@
 /* Copyright (c) 2017 Rod Vagg, MIT License */
 
-var xtend = require('xtend')
-var AbstractIterator = require('./abstract-iterator')
-var AbstractChainedBatch = require('./abstract-chained-batch')
+const xtend = require('xtend')
+const AbstractIterator = require('./abstract-iterator')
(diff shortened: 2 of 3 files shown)


abstract-leveldown.js
```diff
@@ -186,22 +188,41 @@ AbstractLevelDOWN.prototype.batch = function (array, options, callback) {
...
-  options.values = options.values !== false
-  options.limit = 'limit' in options ? options.limit : -1
-  options.keyAsBuffer = options.keyAsBuffer !== false
-  options.valueAsBuffer = options.valueAsBuffer !== false
+function isRangeOption (k) {
+  return rangeOptions.indexOf(k) !== -1
+}
+
+function isEmptyRangeOption (v) {
+  return v === '' || v == null || isEmptyBuffer(v)
```
vweevers: Maybe we should document this?

ralphtheninja: I'll update the PR with docs. Have to run in a few minutes so it'll have to wait until tonight.

test.js
```diff
@@ -950,3 +950,80 @@ test('.status', function (t) {
...
+    const options = {}
+    keys.forEach(function (key) {
+      options[key] = constrFn()
+    })
+    return options
+  }
+
+  function verifyUndefinedOptions (t, options) {
+    keys.forEach(function (key) {
+      t.is(typeof options[key], 'undefined', 'property should be deleted')
```
vweevers: Maybe even do `t.notOk(key in options)`.

README.md
```diff
@@ -114,6 +114,17 @@ By default a `batch()` operation without arguments returns a blank `AbstractChai
...
+
+* `gt`
+* `gte`
+* `lt`
+* `lte`
+* `start` (legacy)
+* `end` (legacy)
+
+A range option that is either an empty buffer, an empty string or `null` will be ignored.
+
```
ralphtheninja: Build for this commit was auto canceled by Travis.

vweevers: It does that? Huh.

ralphtheninja: Yes I just noticed that myself. Actually very useful!

vweevers: Because you pushed another commit, or because only the readme changed?

ralphtheninja: Would be nice though if github could show something else than the red cross, since it's actually a canceled build and you don't know if it would have failed or not.

ralphtheninja: @zeke Can you fix that for me? :grinning:

abstract-leveldown.js
```diff
@@ -186,22 +188,41 @@ AbstractLevelDOWN.prototype.batch = function (array, options, callback) {
 }
 
 AbstractLevelDOWN.prototype._setupIteratorOptions = function (options) {
-  options = xtend(options)
+  var result = cleanRangeOptions(options)
```
vweevers: Nitpick: I prefer `options = cleanRangeOptions(options)`, because the original object isn't (and shouldn't be) used after this line.

ralphtheninja: Of course. I'll fix.

README.md
ralphtheninja: > Because you pushed another commit, or because only the readme changed?

Because I pushed another commit while the first one was building. Apparently it's something you can configure per repository on Travis.

ralphtheninja: auto-cancel

Issue #174, opened 2017-12-23
remove isLevelDOWN utility function?


ralphtheninja:
I added this function but I have no idea what it's used for. I vote for removing it. We can always put it into a module if the usage is spread between several modules.

@juliangruber @vweevers WDYT?

juliangruber:
ftr this commit added it: https://github.com/Level/abstract-leveldown/commit/8051f8f16cb948ac4f858e98a5385fe96a339ad2#diff-a5b095a3840cc9a19a7b58565f942640

this is the PR: https://github.com/Level/abstract-leveldown/pull/64

I'm +-0, I don't know why we need it but also see no harm in keeping it.

ralphtheninja:
Made a search through https://github.com/level/awesome#stores and none of them are using it

vweevers:
I did some quick archeology; correct me if I'm wrong.

This was originally proposed as part of a solution that allowed `levelup(location)` as well as `levelup(db)`. Using `is-leveldown`, levelup would check whether the argument was a db instance. See https://github.com/Level/levelup/pull/299 and https://github.com/Level/levelup/pull/342. It was proposed because at that time you could only pass `db` as a factory function, not as an instance.

But nowadays the signature is `levelup(db)` anyway, where `db` is an instance, no need for `is-leveldown`.

I vote for removing.

PR #178, opened 2018-01-07
remove is-leveldown

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/174


ralphtheninja:
@vweevers After this I'm done and we could release `v4`. :tada:

vweevers:
@ralphtheninja I would like to add #179.

Issue #179, opened 2018-01-07
Update changelog


vweevers:
And it would be nice to have a common approach for Level.

I see the changelog as an upgrade guide for humans. GitHub and git already provide the information in detail; summarizing that is the gap to fill. A commit list in the changelog is redundant, unless you're working offline. Still, online or offline, it often contains too much noise to answer the question "what changed *for me*"? 

One thing the commit list does provide is authorship, or maybe attribution is a better word. If people care about that (and it being available offline), we can opt to have both an upgrade guide and commit list.

I suggest:

- [x] No commit list (but keep old releases as-is)
- [ ] Describe breaking changes, with code examples where needed
- [x] For headers, use `<version> YYYY-MM-DD` for parsability
- [x] Include changelog in npm package (already true, here at least)
- [ ] On release, copy section to GitHub release

WDYT @ralphtheninja?

vweevers:
Or, have both a `CHANGELOG.MD` and `UPGRADING.md`.

ralphtheninja:
I was thinking about this yesterday and it's definitely something that has been lagging behind and we haven't really had any common practices as you say. Spot on! I like the idea of having code examples in the `CHANGELOG.md` file. Not sure we need another file `UPGRADING.md` on top of this though.

vweevers:
> Not sure we need another file UPGRADING.md on top of this though.

Only matters if we want both a commit list and upgrade guide. Let's decide on that first

ralphtheninja:
I want to remove the commit list, it's redundant as you pointed out.

ralphtheninja:
@vweevers Should we keep the current `CHANGELOG.md` as it is? I.e. leave the commit list for previous versions but not adding any new for coming versions?

vweevers:
I think yes. It's better than nothing

ralphtheninja:
@vweevers It's a bit difficult to write something in the `CHANGELOG.md` for a future release, since you don't have the release date (it's in the future, duh). What's your view on what the workflow should look like for updating it?

vweevers:
Add a header "5.0.0 - unreleased" and put new items under there.

ralphtheninja:
And when you add new items you basically update that date later?

vweevers:
We can (strive to) update the changelog whenever we make breaking changes or add features. Then on release, we merely have to replace "unreleased" with the date.

vweevers:
[...]

PR #181, opened 2018-01-07
Update changelog

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/179

~~WIP Don't merge~~

Go through all versions below

- [x] `0.0.0`
- [x] `0.0.1`
- [x] `0.0.2`
- [x] `0.1.0`
- [x] `0.2.0`
- [x] `0.2.1`
- [x] `0.2.2`
- [x] `0.2.3`
- [x] `0.3.0`
- [x] `0.4.0`
- [x] `0.4.0-1`
- [x] `0.5.0`
- [x] `0.6.0`
- [x] `0.6.1`
- [x] `0.7.0`
- [x] `0.7.1`
- [x] `0.7.2`
- [x] `0.7.3`
- [x] `0.7.4`
- [x] `0.8.0`
- [x] `0.8.1`
- [x] `0.8.2`
- [x] `0.9.0`
- [x] `0.10.0`
- [x] `0.10.1`
- [x] `0.10.2`
- [x] `0.11.0`
- [x] `0.11.1`
- [x] `0.11.2`
- [x] `0.11.3`
- [x] `0.11.4`
- [x] `0.12.0`
- [x] `0.12.1`
- [x] `0.12.2`
- [x] `0.12.3`
- [x] `0.12.4`
- [x] `1.0.0`
- [x] `2.0.0`
- [x] `2.0.1`
- [x] `2.0.2`
- [x] `2.0.3`
- [x] `2.1.0`
- [x] `2.1.1`
- [x] `2.1.2`
- [x] `2.1.3`
- [x] `2.1.4`
- [x] `2.2.0`
- [x] `2.2.1`
- [x] `2.2.2`
- [x] `2.3.0`
- [x] `2.3.1`
- [x] `2.4.0`
- [x] `2.4.1`
- [x] `2.5.0 `
- [x] `2.6.0`
- [x] `2.6.1`
- [x] `2.6.2`
- [x] `2.6.3`
- [x] `2.7.0`
- [x] `2.7.1`
- [x] `2.7.2`
- [x] `3.0.0`
- [ ] ~~add upgrade guide to coming v4 release, i.e. `Unreleased` section~~
- [x] use `Added`, `Changed`, `Removed` and `Fixed` appropriately for each version, see https://github.com/olivierlacan/keep-a-changelog/blob/master/CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -243,63 +243,54 @@ No change.
 ## [0.12.4] - 2014-08-20
 
 ### Added
+* test that `simple-iterator` returns buffers (@kesla)
+* test for implicit snapshots (@kesla)
+
 ### Changed
-### Fixed
-### Removed
+* license to plain MIT (@andrewrk)
 
 ## [0.12.3] - 2014-06-27
 
-### Added
 ### Changed
-### Fixed
-### Removed
+* Update `xtend` dependency (@andrewrk)
 
 ## [0.12.2] - 2014-04-26
 
-### Added
 ### Changed
-### Fixed
-### Removed
+* `isTypedArray` checks for existence of `ArrayBuffer` and `Uint8Array` constructors before usage (@rvagg)
 
 ## [0.12.1] - 2014-04-26
 
-### Added
 ### Changed
-### Fixed
-### Removed
+* `BufferType` in `abstract/put-get-del-test.js` defaults to `Buffer` instead of `ArrayBuffer` (@maxogden)
 
 ## [0.12.0] - 2014-03-12
 
-### Added
 ### Changed
-### Fixed
-### Removed
+* Revert to pure `Buffer` and remove usage of `Uint16Array` (@rvagg)
 
 ## [0.11.4] - 2014-03-11
 
-### Added
-### Changed
-### Fixed
 ### Removed
+* Remove duplicate call to `t.end()` (@maxogden)
 
 ## [0.11.3] - 2014-01-26
 
-### Added
 ### Changed
-### Fixed
-### Removed
+* Less restrictive check for buffer type (@rvagg)
 
 ## [0.11.2] - 2013-12-05
 
 ### Added
-### Changed
+* npm badges (@rvagg)
+
 ### Fixed
-### Removed
+* broken iterator tests in `test.js` (@rvagg)
 
 ## [0.11.1] - 2013-11-15
 
 ### Changed
-* Adjust approximate-size-test.js to account for snappy compression
+* Adjust `abstract/approximate-size-test.js` to account for snappy compression (@rvagg)
 
 ## [0.11.0] - 2013-10-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,145 +61,113 @@
 ## [2.5.0] - 2016-05-01
 
 ### Added
+* dependency badge to `README` (@ralphtheninja)
+* `AbstractLevelDOWN#_serializeKey` (@juliangruber)
+* `AbstractLevelDOWN#_serializeValue` (@juliangruber)
+* `AbstractChainedBatch#_serializeKey` (@juliangruber)
+* `AbstractChainedBatch#_serializeValue` (@juliangruber)
+* tests for `_serialize` object and buffer (@juliangruber)
+
 ### Changed
+* Update dependencies and add more node versions to Travis (@ralphtheninja)
+
 ### Fixed
+* Update `memdown` url (@ralphtheninja)
+* `AbstractLevelDOWN#._checkKey` does not take three parameters (@ralphtheninja)
+* Only show build status for the master branch (@watson)
+* Fix minor typos in `README` (@timkuijsten)
+
 ### Removed
-* [[`fd4ba39136`](https://github.com/level/abstract-leveldown/commit/fd4ba39136)] - 2.5.0 (@juliangruber)
-* [[`b00f97f9ac`](https://github.com/level/abstract-leveldown/commit/b00f97f9ac)] - Merge pull request #85 from Level/remove/stringify-key-value (Lars-Magnus Skog)
-* [[`1138d6d993`](https://github.com/level/abstract-leveldown/commit/1138d6d993)] - make tests more specific (@juliangruber)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -19,109 +23,136 @@
...
+[2.2.0]: https://github.com/level/abstract-leveldown/compare/v2.1.4...v2.2.0
+[2.1.4]: https://github.com/level/abstract-leveldown/compare/v2.1.3...v2.1.4
+[2.1.3]: https://github.com/level/abstract-leveldown/compare/v2.1.2...v2.1.3
+[2.1.2]: https://github.com/level/abstract-leveldown/compare/v2.1.1...v2.1.2
+[2.1.1]: https://github.com/level/abstract-leveldown/compare/v2.1.0...v2.1.1
+[2.1.0]: https://github.com/level/abstract-leveldown/compare/v2.0.3...v2.1.0
+[2.0.3]: https://github.com/level/abstract-leveldown/compare/v2.0.2...v2.0.3
+[2.0.2]: https://github.com/level/abstract-leveldown/compare/v2.0.1...v2.0.2
+[2.0.1]: https://github.com/level/abstract-leveldown/compare/v2.0.0...v2.0.1
+[2.0.0]: https://github.com/level/abstract-leveldown/compare/0.11.1...v2.0.0
```
ralphtheninja: I'm aware of the big gap here. Will adjust later once I have added the missing versions/tags.

ralphtheninja:
@vweevers I just noticed something odd in the git history of `abstract-leveldown`, see screenshot below

weird-git-history

There's a remote `patch-1` branch which contains tags `v1.1.0` to `v1.8.1` which does _not_ correspond to any existing versions of `abstract-leveldown`. We only released `v1.0.0` before jumping directly to `v2.0.0`.

ralphtheninja:
Seems to be releated to `AbstractNoSQL`. I'm wondering if @snowyu accidentally pushed to the wrong remote or something like that.

vweevers:
😮 If we don't get a timely response from @snowyu, let's delete that branch.

CHANGELOG.md
```diff
@@ -151,8 +287,32 @@ No change.
...
+[2.0.0]: https://github.com/level/abstract-leveldown/compare/v1.0.0...v2.0.0
+[1.0.0]: https://github.com/level/abstract-leveldown/compare/v0.12.4...v1.0.0
+[0.12.4]: https://github.com/level/abstract-leveldown/compare/v0.12.3...v0.12.4
+[0.12.3]: https://github.com/level/abstract-leveldown/compare/v0.12.2...v0.12.3
+[0.12.2]: https://github.com/level/abstract-leveldown/compare/v0.12.1...v0.12.2
+[0.12.1]: https://github.com/level/abstract-leveldown/compare/v0.12.0...v0.12.1
+[0.12.0]: https://github.com/level/abstract-leveldown/compare/v0.11.4...v0.12.0
+[0.11.4]: https://github.com/level/abstract-leveldown/compare/v0.11.3...v0.11.4
+[0.11.3]: https://github.com/level/abstract-leveldown/compare/v0.11.2...v0.11.3
+[0.11.2]: https://github.com/level/abstract-leveldown/compare/0.11.1...v0.11.2
```
ralphtheninja: Early tags don't start with a `v` which is a bit annoying.

ralphtheninja:
> If we don't get a timely response from @snowyu, let's delete that branch.

@vweevers I pushed up a branch named `THIS_BRANCH_WILL_BE_DELETED` and removed all the tags. If you have any local tags, make sure you don't do `git push --tags` or similar.

ralphtheninja:
@vweevers Check current state at https://github.com/Level/abstract-leveldown/blob/changelog/CHANGELOG.md

ralphtheninja:
A suggestion is to use links for all contributors, e.g. @vweevers would be wrapped with a link to your github page.

ralphtheninja:
@vweevers Ok, done with `Unreleased` as well. *wipes sweat from his forehead* :grinning: I think this changelog is starting to look pretty bad ass actually.

vweevers:
Perfect. This was done with care, with great attention to detail and consistency. 

I fully agree a changelog in this form is valuable to have, separate from a upgrade guide yet more human than commit logs.

This serves as a blueprint for all changelogs. I love the little Historical Notes. Acknowledging mistakes goes a long way.

Couple of questions:

- What tools did you use for this?
- Should we make an effort to write better commit messages, so we can automate the changelog (more)?
- Which commits did you exclude? Any recommendations we can follow in the future?

vweevers:
> Should we make an effort to write better commit messages

Better meaning more consistent and to the point.

ralphtheninja:
> What tools did you use for this?

Pure manual labor and emacs :)

> Should we make an effort to write better commit messages, so we can automate the changelog (more)?

I thought about this. There's a bunch of trade offs to be made here. If we go down the path of forcing a certain type of commit messages, then changelog automation is trivial ([generate-changelog](https://github.com/lob/generate-changelog) does this), but once in a while a commit message will slip through that doesn't follow the rules and then automation will be problematic. Also, I'm not sure it's beneficial to enforce a certain commit style, for the sole purpose of automation. That said, we could always try to write better commit messages (i.e. not "changed some stuff") even though they aren't following a predetermined format.

Maybe it's better to have a simple tool that just checks the current changelog and makes sure it has an entry for respective release. E.g. lets say that I create a new release (new tag and new npm release) but forget to write an entry in the changelog, then it would be trivial to have tool check for this entry, and have tests fail or something like that.

Once a changelog exists, it's not a lot of work to either update it at each release, or update it iteratively as new commits come in.

> Which commits did you exclude? Any recommendations we can follow in the future?

I went through the whole history and all the diffs and while doing this I've noticed commits that undoes previous commits. For example, someone commits a change, then gets feedback and makes more changes that removes or modifies the previous commit. I decided to not document this, but rather document the final state of the contribution.

ralphtheninja:
@vweevers I had an idea the other day about a `changelogify` module, which could help out with either generating a changelog template, i.e. grab versions from npm and get tags from git, and use that information to generate a starting changelog that has all the fields filled in. It could generate all the `compare`-links and whatnot.

Another idea could be to create `verify-changelog` that could be used as a testing tool.

Issue #185, opened 2018-01-10
Add UPGRADING.md


ralphtheninja:
@vweevers What would the format look like here?

ralphtheninja:
Once this issue is resolved, I vote for a new release.

vweevers:
Sorry for the late response. I have no preference for the format, and don't know of a good convention.

PR #188, opened 2018-01-19
add UPGRADING.md

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/185

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,7 @@
 # Upgrade Guide
 
+This document describes breaking changes and how to upgrade. For a complete list of changes including minor and patch releases, please refer to the changelog.
+
 ## v4
 
 #### default `testCommon` parameter
@@ -68,9 +70,7 @@ abstract.all(leveldown, test)
 
 #### `.approximateSize` method removed
 
-The `.approximateSize` method has been removed. The rationale behind this is that it's a feature heavily related to `LevelDB` and not necessarily something that's used in other stores. If your store has implemented `._approximateSize` it will never be called by `abstract-leveldown`.
-
-If your implementation needs this, it must rename `._approximateSize` to `.approximateSize` and also take care of the initialization code, take a look at what `leveldown` has done if you need more details.
+The `.approximateSize` method has been removed from the public API. It is heavily related to `LevelDB` and more often than not, other stores lack the native primitives to implement this. If you did implement the internal `_approximateSize` method, that is now dead code. To preserve the method in your public API, rename it to `approximateSize` and also take care of the initialization code. Look to `leveldown` for inspiration.
 
 Also, the corresponding abstract tests have been removed, so your implementation can no longer require `abstract/approximate-size-test`.
 
@@ -80,7 +80,7 @@ Because `Buffer` is available in all environments nowadays, there is no need for
 
 #### `isLevelDOWN` function removed
 
-This was a legacy function and no implementation seems to use it.
+This was a legacy function.
 
 #### `ranges-test.js` renamed
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -2,11 +2,9 @@
 
 ## v4
 
-#### `testCommon` parameter defaults
+#### default `testCommon` parameter
 
-The `testCommon` variable will now default to `testCommon.js` in `abstract-leveldown`. Unless your implementation needs a modified version, you can just omit the parameter.
-
-For instance, `leveldown` requires `testCommon.js` directly from `abstract-leveldown` and passes it in as a parameter, which makes for unnecessary boiler plate.
+The `testCommon` parameter will now default to `abstract-leveldown/testCommon.js`. You can omit this parameter, unless your implementation needs a custom version.
 
 If your code today looks something like:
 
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -0,0 +1,105 @@
+# Upgrade Guide
+
+## v4
+
+#### `testCommon` parameter defaults
+
+The `testCommon` variable will now default to `testCommon.js` in `abstract-leveldown`. Unless your implementation needs a modified version, you can just omit the parameter.
+
+For instance, `leveldown` requires `testCommon.js` directly from `abstract-leveldown` and passes it in as a parameter, which makes for unnecessary boiler plate.
```
vweevers: I think you can leave this sentence out. The example below says it all.

UPGRADING.md
```diff
@@ -0,0 +1,105 @@
...
+const abstract = require('abstract-leveldown/abstract/get-test')
+
+abstract.all(leveldown, test)
+```
+
+#### `testBuffer` parameter removed
+
+The `abstract/put-get-del-test.js` previously took a custom `testBuffer` parameter, which now has been removed. After an [analysis](https://github.com/Level/abstract-leveldown/pull/175#issuecomment-353867144) of different implementations we came to the conclusion that no implementation really needed this parameter. They basically just passed in a `Buffer` with a custom value which is more unnecessary boiler plate.
+
+If your implementation are using this abstract test, change from:
```
vweevers: - are → is

UPGRADING.md
```diff
@@ -0,0 +1,105 @@
...
+
+#### `isLevelDOWN` function removed
+
+This was a legacy function and no implementation seems to use it.
+
+#### `ranges-test.js` renamed
+
+We have refactored a lot of the tests, specifically the iterator tests were split into two tests and in that process we renamed `ranges-test.js` to `iterator-range-test.js`.
+
+If your implementation are using these tests then change from:
```
vweevers: - are → is

UPGRADING.md
```diff
@@ -0,0 +1,105 @@
...
+
+If you rely on this use `Buffer.isBuffer()` instead.
+
+#### `isLevelDOWN` function removed
+
+This was a legacy function and no implementation seems to use it.
+
+#### `ranges-test.js` renamed
+
+We have refactored a lot of the tests, specifically the iterator tests were split into two tests and in that process we renamed `ranges-test.js` to `iterator-range-test.js`.
```
vweevers: - "We have refactored a lot of the tests. Specifically, "

UPGRADING.md
```diff
@@ -0,0 +1,105 @@
...
+const test = require('tape')
+const leveldown = require('./your-leveldown')
+const abstract = require('abstract-leveldown/abstract/get-test')
+
+abstract.all(leveldown, test)
+```
+
+#### `testBuffer` parameter removed
+
+The `abstract/put-get-del-test.js` previously took a custom `testBuffer` parameter, which now has been removed. After an [analysis](https://github.com/Level/abstract-leveldown/pull/175#issuecomment-353867144) of different implementations we came to the conclusion that no implementation really needed this parameter. They basically just passed in a `Buffer` with a custom value which is more unnecessary boiler plate.
```
vweevers: - Remove redundant "which has now been removed"
- different implementations → various implementations
- Simply ".. we came to the conclusion that the parameter has no use [end of paragraph.]" The analysis you linked to has all the details.

UPGRADING.md
```diff
@@ -0,0 +1,105 @@
...
+
+The `.approximateSize` method has been removed. The rationale behind this is that it's a feature heavily related to `LevelDB` and not necessarily something that's used in other stores. If your store has implemented `._approximateSize` it will never be called by `abstract-leveldown`.
+
+If your implementation needs this, it must rename `._approximateSize` to `.approximateSize` and also take care of the initialization code, take a look at what `leveldown` has done if you need more details.
+
+Also, the corresponding abstract tests have been removed, so your implementation can no longer require `abstract/approximate-size-test`.
+
+#### `._isBuffer` method removed
+
+If you rely on this use `Buffer.isBuffer()` instead.
```
vweevers: Could use some context here. Suggestion: "Because `Buffer` is available in all environments nowadays, there is no need for alternatives like typed arrays. It is preferred to use `Buffer` and `Buffer.isBuffer()` directly."

UPGRADING.md
ralphtheninja: Also I think `split into two tests` -> `split in two`.

ralphtheninja:
@vweevers Thanks for your feedback. I just threw something together in the middle of the night.

UPGRADING.md
```diff
@@ -0,0 +1,103 @@
+# Upgrade Guide
+
```
vweevers: Maybe add an intro like "This document describes breaking changes [and how to upgrade?]. For a complete list of changes including minor and patch releases, please refer to the changelog".

UPGRADING.md
```diff
@@ -0,0 +1,103 @@
...
+const test = require('tape')
+const leveldown = require('./your-leveldown')
+const abstract = require('abstract-leveldown/abstract/put-get-del-test')
+
+abstract.all(leveldown, test)
+```
+
+#### `.approximateSize` method removed
+
+The `.approximateSize` method has been removed. The rationale behind this is that it's a feature heavily related to `LevelDB` and not necessarily something that's used in other stores. If your store has implemented `._approximateSize` it will never be called by `abstract-leveldown`.
```
vweevers: I think we need to clarify the distinction between public and internal. Suggestion with some additional tweaks:

"The `approximateSize` method has been removed from the public API. It is heavily related to `LevelDB` and more often than not, other stores lack the native primitives to implement this. If you did implement the internal `_approximateSize` method, that is now dead code. To preserve the method in your public API, rename it to `approximateSize` and also take care of the initialization code. Look to `leveldown` for inspiration."

UPGRADING.md
```diff
@@ -0,0 +1,103 @@
...
+
+Also, the corresponding abstract tests have been removed, so your implementation can no longer require `abstract/approximate-size-test`.
+
+#### `._isBuffer` method removed
+
+Because `Buffer` is available in all environments nowadays, there is no need for alternatives like typed arrays. It is preferred to use `Buffer` and `Buffer.isBuffer()` directly.
+
+#### `isLevelDOWN` function removed
+
+This was a legacy function and no implementation seems to use it.
```
vweevers: I think "This was a legacy function." would suffice. Even if some implementation in a dark corner of the internet does use it, doesn't mean it should.

ralphtheninja:
`4.0.0`

ralphtheninja:
@juliangruber Feel like tweeting it out? I'll toot it on Mastodon.

Issue #193, opened 2018-01-22
Consider defining default methods on prototype


vweevers:
To make TypeScript stop complaining about undefined types, instead of:

```js
AbstractChainedBatch.prototype.put = function (key, value) {
  // ..

  // @ts-ignore
  if (typeof this._put === 'function') {
    // @ts-ignore
    this._put(key, value)
  } else {
    this._operations.push({ type: 'put', key: key, value: value })
  }

  // ..
}
```

We could write:

```js
AbstractChainedBatch.prototype.put = function (key, value) {
  // ..

  this._put(key, value)

  // ..
}

AbstractChainedBatch.prototype._put = function (key, value) {
  this._operations.push({ type: 'put', key: key, value: value })
}
```

Same with all the other methods. Thoughts @ralphtheninja @juliangruber?

vweevers:
@ralphtheninja such a change could be a patch release, right? It doesn't change behavior, unless someone is doing some funky prototype stuff (I wouldn't know what).

ralphtheninja:
Aye, a patch version should be sufficient.

vweevers:
We have a weird one here:

https://github.com/Level/abstract-leveldown/blob/0d942fd45df5c9915269ded6f0f76fdfade11575/abstract-chained-batch.js#L79-L85

Notice that `_write` doesn't receive `options`.

Issue #192, opened 2018-01-21
Remove unused testCommon parameter


vweevers:
Signature:

https://github.com/Level/abstract-leveldown/blob/0d942fd45df5c9915269ded6f0f76fdfade11575/abstract/leveldown-test.js#L1

Call:

https://github.com/Level/abstract-leveldown/blob/0d942fd45df5c9915269ded6f0f76fdfade11575/test.js#L17

vweevers:
(I'm not discovering these things, TypeScript is)

ralphtheninja:
@vweevers Can we run this automatically somehow during tests?

vweevers:
Yea, I'm working on running the full suite with `ts-node`.

vweevers:
@ralphtheninja done in #195.

PR #195, opened 2018-01-22
Add and fix TypeScript tests

### LINKED ISSUE #190: Update typescript definitions
- Remove `approximateSize`
- Remove  `isLevelDOWN`
- Anything else?

I hope v4 didn't break things for typescript users. @Level/typescript, cc @ralphtheninja

vweevers:
Closes #190, #192, #193 (with a few leftover tasks I'll open issues for).

### DIFF
--- a/abstract-iterator.js
+++ b/abstract-iterator.js
@@ -22,17 +22,14 @@ AbstractIterator.prototype.next = function (callback) {
   }
 
   self._nexting = true
-  if (typeof self._next === 'function') {
-    return self._next(function () {
-      self._nexting = false
-      callback.apply(null, arguments)
-    })
-  }
-
-  process.nextTick(function () {
+  self._next(function () {
     self._nexting = false
-    callback()
+    callback.apply(null, arguments)
   })
+}
+
+AbstractIterator.prototype._next = function (callback) {
+  process.nextTick(callback)
 }
 
 AbstractIterator.prototype.end = function (callback) {
@@ -45,9 +42,10 @@ AbstractIterator.prototype.end = function (callback) {
   }
 
   this._ended = true
+  this._end(callback)
+}
 
-  if (typeof this._end === 'function') { return this._end(callback) }
-
+AbstractIterator.prototype._end = function (callback) {
   process.nextTick(callback)
 }
 
--- a/test-types.ts
+++ b/test-types.ts
@@ -1,0 +1,21 @@
+import { AbstractLevelDOWN } from '.'
+
+const buf = Buffer.from('foo')
+
+// Key and value generics
+new AbstractLevelDOWN<string, string>('loc').put('key', 'value', (err: Error) => {})
+new AbstractLevelDOWN<Buffer, Buffer>('loc').put(buf, buf, (err: Error) => {})
+new AbstractLevelDOWN<number, object>('loc').put(1, {}, (err: Error) => {})
+
+new AbstractLevelDOWN<string, string>('loc').get('key', (err: Error, value: string) => {})
+new AbstractLevelDOWN<Buffer, Buffer>('loc').get(buf, (err: Error, value: Buffer) => {})
+new AbstractLevelDOWN<number, object>('loc').get(1, (err: Error, value: object) => {})
+
+// The generics default to "any"
+new AbstractLevelDOWN('loc').put(1, {}, (err: Error) => {})
+new AbstractLevelDOWN('loc').put('key', 'value', (err: Error) => {})
+new AbstractLevelDOWN('loc').put(buf, false, (err: Error) => {})
+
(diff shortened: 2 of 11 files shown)


vweevers:
(We need more @Level/typescript people)

abstract-iterator.js
```diff
@@ -22,19 +22,16 @@ AbstractIterator.prototype.next = function (callback) {
...
-      self._nexting = false
-      callback.apply(null, arguments)
-    })
-  }
-
-  process.nextTick(function () {
+  self._next(function () {
     self._nexting = false
-    callback()
+    callback.apply(null, arguments)
```
ralphtheninja: Nice one. Much more cleaner now.

vweevers:
I'm thinking we should mark the typings as `stability: experimental` here too.

vweevers:
> I'd like to suggest we add a test.ts to test TypeScript more natively.

@zixia what are the benefits? And does it involve rewriting the test suite in TypeScript?

zixia:
@vweevers The benefits would be able to do static type checking? I do not know whether ts-node do type checking when we run a `.js` file.

After ts-node v4, we have to do a `ts-node --type-check` because the default behavior had been changed.

> And does it involve rewriting the test suite in TypeScript?

I think we can just add one unit test file written by TypeScript, and we use `ts-node --type-check test.ts` to run the test under TypeScript with static type checkings.

vweevers:
> The benefits would be able to do static type checking? I do not know whether ts-node do type checking when we run a .js file.

It does (though to a lesser extent, I'm assuming), which is why we need most commits here (b1d006eab4f0665b471b407ac93d62e60757442b for example).

> After ts-node v4, we have to do a ts-node --type-check because the default behavior had been changed.

Ah! I was wondering why v4 didn't work. Great, I'll upgrade `ts-node` to v4.

vweevers:
Remaining tasks:

- [ ] upgrade `ts-node`
- [ ] add stability badge to readme
- [ ] add simple `test.ts`, run it in addition to `test.js`

vweevers:
@zixia is there a way to test the types themselves? For example, this should be a failing test:

```ts
const db = new AbstractLevelDOWN<string, number>('foobar')

db.put('key', 'value', (err) => {
  t.ifError(err)
  t.end()
})
```

While this one should pass:

```ts
const db = new AbstractLevelDOWN<string, number>('foobar')

db.put('key', 2, (err) => {
  t.ifError(err)
  t.end()
})
```

Without tests like this, I don't know how to make `test.ts` valuable.

vweevers:
Or in other words: how can we test that the type definitions are correct?

test-types.ts
```diff
@@ -0,0 +1,21 @@
+import { AbstractLevelDOWN } from '.'
+
+const buf = Buffer.from('foo')
+
+// Key and value generics
+new AbstractLevelDOWN<string, string>('loc').put('key', 'value', (err: Error) => {})
```
vweevers: Note this isn't that valuable, as it's only positive tests, no negative tests.

vweevers:
PTAL

test-types.ts
vweevers: Also I initially created this file as `test.ts` but this tripped up TypeScript's module resolution (because we also have `test.js`). So I renamed it to `test-types.ts`.

MeirionHughes:
one thing I just couldn't get right was the `asBuffer` or `keyAsBuffer` option stuff. 

i.e. when `K=string` and `options={keyAsBuffer:true}` then the typing for the key on 
```
get(key: K, options: GO, callback: (err: any, value: V) => any): void;
```
is wrong.  Some other solutions I tried, but didn't really work when it came to inference at the levelup stage: 

manually override when needed:
```
get<VV = V>(key: K, options: GO, callback: (err: any, value: VV) => any): void;
```

overloading: 
```
get(key: K, options: GO, callback: (err: any, value: V) => any): void;
get(key: K, options: GO & {valueAsBuffer:true}, callback: (err: any, value: Buffer) => any): void;
```

vweevers:
@MeirionHughes Considering the dynamic nature of everything put together (`abstract-leveldown`, `*asBuffer` options, stores, serialization, `encoding-down`, sublevels, per-operation encodings, etc), using `any` instead of generics does sound like a good idea to me.

It would simplify integration, decrease coupling, make it easier to maintain for the non-TS-savvy, and we could also increase correctness of the rest of the typings (for example, the `*asBuffer` options should theoretically be defined here, not in implementations, as it is `abstract-leveldown` that defines these options and their default values).

That said, if you can find a way to make generics work, then great, go for it ;) Perhaps continue the discussion in a new ticket with @Level/typescript.

As for this PR, I'm gonna squash a few commits and then we're good to merge. I'm not really happy with having to use `class` in tests, as we don't use classes anywhere else, but it's an acceptable trade-off IMO.

PR #199, opened 2018-02-08
Update changelog with TypeScript changes

vweevers:


vweevers:
> Would you mind fixing all capital letters for bulleted lists for the rest of the file?

Sure.

vweevers:
@ralphtheninja on a related note, should we start all lines with a verb (e.g. "Add x", "Remove y")?

ralphtheninja:
> @ralphtheninja on a related note, should we start all lines with a verb (e.g. "Add x", "Remove y")?

Yeah, lets do that. I was thinking about it before. At least whenever it makes sense.

vweevers:
@ralphtheninja for adding a release title and date to the changelog, is it OK to push to master? I'm not sure what the workflow for publishing should be.

ralphtheninja:
@vweevers Yes I think so. When publishing I normally do `npm version major|minor|patch` and push to master and push up the tags. Now fixing the changelog is basically part of that workflow, so entirely ok since it's a trivial change.

vweevers:
4.0.1

PR #202, opened 2018-02-09
Fix iterator#next to return this

vweevers:
Fixes https://github.com/Level/leveldown/issues/443#issuecomment-364437129. I also updated the TypeScript typings @Level/typescript.

Should be released asap.

### DIFF
--- a/abstract-iterator.js
+++ b/abstract-iterator.js
@@ -14,11 +14,13 @@ AbstractIterator.prototype.next = function (callback) {
   }
 
   if (self._ended) {
-    return process.nextTick(callback, new Error('cannot call next() after end()'))
+    process.nextTick(callback, new Error('cannot call next() after end()'))
+    return self
   }
 
   if (self._nexting) {
-    return process.nextTick(callback, new Error('cannot call next() before previous next() has completed'))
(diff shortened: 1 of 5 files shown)


vweevers:
I'm testing it against leveldown.

abstract-iterator.js
```diff
@@ -14,18 +14,22 @@ AbstractIterator.prototype.next = function (callback) {
...
+    return self
   }
 
   self._nexting = true
   self._next(function () {
     self._nexting = false
     callback.apply(null, arguments)
   })
+
+  return self
```
ralphtheninja: Aah it was chaining that was the problem.

abstract/iterator-test.js
```diff
@@ -40,6 +40,13 @@ module.exports.args = function (test) {
     )
     iterator.end(t.end.bind(t))
   })
+
+  test('test iterator#next returns this', function (t) {
+    var iterator = db.iterator()
```
ralphtheninja: 👍

vweevers:
leveldown is passing.

vweevers:
@ralphtheninja good to release?

ralphtheninja:
Yep. Go!

PR #204, opened 2018-02-19
Fix typings for Batch

Tapppi:
Tried using the typings with typescript 2.6/2.7 and neither worked. Both throw on `type Batch<K=any, V?=any>`. https://github.com/Level/abstract-leveldown/commit/b353fb45b00165222aa616111ddc8ae00e9a860a added the `?` before the default for some reason. AFAIK this is not valid typescript, and servers no purpose. Since `type Batch<K=any, V=any>` already has defaults, the values are optional. Without the question mark it seems to be working.

@vweevers can you elaborate on what you were trying to achieve?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,7 +40,7 @@ export interface AbstractIteratorOptions<K=any> {
   values?: boolean;
 }
 
-export type Batch<K=any, V?=any> = PutBatch<K, V> | DelBatch<K>
+export type Batch<K=any, V=any> = PutBatch<K, V> | DelBatch<K>
 
 export interface PutBatch<K=any, V=any> {
   type: 'put',
--- a/test.js
+++ b/test.js
@@ -230,6 +230,7 @@ test('test batch() extensibility', function (t) {
   var spy = sinon.spy()
   var expectedCb = function () {}
   var expectedOptions = { options: 1 }
+  /** @type {Array<{ type: 'put', key, value } | { type: 'del', key }>} */
   var expectedArray = [
     { type: 'put', key: '1', value: '1' },
     { type: 'del', key: '2' }
@@ -676,6 +677,7 @@ test('test serialization extensibility (batch array is not mutated)', function (
 
   test = new Test('foobar')
 
+  /** @type { { type: 'put', key, value } } */
   var op = { type: 'put', key: 'no', value: 'nope' }
   test.batch([op], function () {})
 


vweevers:
Without it, `npm run test-ts` failed for me. I don't recall why, and it's best if someone with TS knowledge takes a look at this. @Level/typescript

Tapppi:
> Without it, npm run test-ts failed for me. I don't recall why, and it's best if someone with TS knowledge takes a look at this

Yep, noticed, I'll try to see what's up but it's a pretty mystic failure..

Tapppi:
Okay, figured out what is wrong.. The type for Batch:
```
export type Batch<K=any, V=any> = PutBatch<K, V> | DelBatch<K>

export interface PutBatch<K=any, V=any> {
  type: 'put',
  key: K,
  value: V
}

export interface DelBatch<K=any> {
  type: 'del',
  key: K
}
```

The test input that results in error, even though there should be nothing wrong:
```
var expectedArray: {}[] = [
    { type: 'put', key: '1', value: '1' },
    { type: 'del', key: '2' }
  ]
// Should be compatible with Batch<any, any>[], but produces error:
test.js (244,14): Argument of type '({ [x: string]: any; type: string; key: string; value: string; } | { [x: string]: any; type: stri...' is not assignable to parameter of type 'Batch<any, any>[]'.
  Type '{ [x: string]: any; type: string; key: string; value: string; } | { [x: string]: any; type: strin...' is not assignable to type 'Batch<any, any>'.
    Type '{ [x: string]: any; type: string; key: string; value: string; }' is not assignable to type 'Batch<any, any>'.
      Type '{ [x: string]: any; type: string; key: string; value: string; }' is not assignable to type 'DelBatch<any>'.
        Types of property 'type' are incompatible.
          Type 'string' is not assignable to type '"del"'. (2345)
```

This is something I did hit in my own code at one point when using discriminated unions. For some reason the object `{ type: 'del', key: '2' }` is getting inferred as `{ ... ,type: 'string, ...}` which does not pass the type check. @Level/typescript if anyone knows how to deal with this in a nice way, I would be interested. But as it stands, this is a problem in inferring the test types and the current typings in master are plain invalid. 

@vweevers Why didn't the tests turn red for invalid typings? If the typings file doesn't compile at all, it is ignored instead of throwing an error?

vweevers:
Thanks for looking into this!

```js
var expectedArray = [
  { type: 'put', key: '1', value: '1' },
  { type: 'del', key: '2' }
]
```

Is there a way to annotate JS like this, perhaps with JSDoc, to cast it to the right type? So that we can get our tests passing and land this PR.

vweevers:
> @vweevers Why didn't the tests turn red for invalid typings? If the typings file doesn't compile at all, it is ignored instead of throwing an error?

That would be bad. TypeScript not barking led me to believe it was valid (and that it did compile). If you're right about TS ignoring the typings altogether, I hope there's a way to disable that behavior?

It was also my mistake, thinking I could do TS. Would you be interested in helping out to maintain the typings?

Tapppi:
> It was also my mistake, thinking I could do TS. Would you be interested in helping out to maintain the typings?

I'm fairly new to the world of TS myself, but am invested in it as I'll be using TS and `level-rocksdb` for a major Kafka Streams TS project. My time is fairly contested right now, but should be able to try and help with the typings. I'm taking another stab at this as I think I'll be needing this solved for my own projects as well..

Tapppi:
If I put this clearly invalid typescript into the typings, all `npm run test-ts` tests run fine.
```
asfdasdf
export ???ASDFASDF???? type Batch<K=any, V?=any> = PutBatch<K, V> | DelBatch<K>
```
If I fix the typings, the tests bark at the invalid typecheck. This is fun 😄 

Apparently `node-ts` doesn't actually check that it get's types, but ignores the malformed file. It just checks types when it gets the correct typings and then reacts to the type error. `tsc -p .` (Typescript compiler) *does* bark at the malformed typings.

>  If you're right about TS ignoring the typings altogether, I hope there's a way to disable that behavior?

Don't see a way of disabling this behaviour in `ts-node`.

> Is there a way to annotate JS like this, perhaps with JSDoc, to cast it to the right type? So that we can get our tests passing and land this PR.
Test file isn't TS so there isn't a way to use typings without using `.ts` extension (at least to my knowledge?). 

If I convert to `test.ts`, I can fix the error by typing the batch array explicitly:
```
var expectedArray: {type: 'put' | 'del', key: string, value?: string}[] = [
    { type: 'put', key: '1', value: '1' },
    { type: 'del', key: '2' }
  ]
// or just:
var expectedArray: Batch[] = [
    { type: 'put', key: '1', value: '1' },
    { type: 'del', key: '2' }
  ]
```
But theres a *ton* of typescript compiler errors in test.ts, it's pretty loosely (javascriptly) written..

zixia:
I had cloned the @Tapppi version of this PR in my Ubuntu box, and it passed all the tests by `npm i && npm t` without any problem. 

Strange.

Tapppi:
I'm thinking this is some weird problem with `ts-node`. If I run `npm run test-ts` I still get the errors with this branch..

I got all of the problems fixed by converting the test file to typescript (`test.ts`), explicitly setting all constructor types and the `Batch[]` typing. Then used `tsc -p .` to compile and type check. Not sure what the consensus would be on doing that.. Also don't know if the explicit constructor types is something that will bug out with dependant libraries as well. Might have to test that out.

@zixia can you please try renaming the test file to `test.ts` and running:
```
[yarn|npx] ts-node -v
[yarn|npx] tsc test2.ts index.d.ts global.d.ts test-types.ts

// My versions:
ts-node v4.1.0
node v6.12.2
typescript v2.7.2
```

Tried with node 8 too, but that made no difference. If we still get diverging behaviour this is really weird. If we don't it's probably `ts-node` fucking things up.

zixia:
```shell
$ ./node_modules/.bin/ts-node -v
ts-node v4.1.0
node v9.5.0
typescript v2.7.2

$ ./node_modules/.bin/tsc test.ts index.d.ts global.d.ts test-types.ts
test.ts(55,14): error TS2554: Expected 0 arguments, but got 1.
test.ts(64,14): error TS2554: Expected 0 arguments, but got 1.
test.ts(95,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(105,19): error TS2339: Property 'options' does not exist on type '{ createIfMissing: boolean; errorIfExists: boolean; }'.
test.ts(122,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(141,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(153,19): error TS2339: Property 'options' does not exist on type '{ asBuffer: boolean; }'.
test.ts(174,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(206,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(242,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(282,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(316,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(334,14): error TS2554: Expected 0 arguments, but got 1.
test.ts(348,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(373,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(393,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(411,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(436,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(448,14): error TS2554: Expected 0 arguments, but got 1.
test.ts(461,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(483,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(513,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(540,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(569,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(597,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(624,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(650,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(677,10): error TS2554: Expected 0 arguments, but got 1.
test.ts(696,12): error TS2554: Expected 0 arguments, but got 1.
test.ts(720,12): error TS2554: Expected 0 arguments, but got 1.
test.ts(737,12): error TS2554: Expected 0 arguments, but got 1.
test.ts(756,12): error TS2554: Expected 0 arguments, but got 1.
test.ts(774,12): error TS2554: Expected 0 arguments, but got 1.
```

Aha, I got the errors...

zixia:
I believe we have to put `as any` to our mocked data, as the following:
```js
  var expectedArray = [
    { type: 'put', key: '1', value: '1' },
    { type: 'del', key: '2' }
  ] as any
```

Or better to set it to `as Batch<any, any>[]`

However, this could not be possible if we name the file `test.js`, we have to rename the `test.js` to `test.ts` to satisfied this change.

vweevers:
> However, this could not be possible if we name the file test.js, we have to rename the test.js to test.ts to satisfied this change.

Perhaps try https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript to annotate the JS. We have to keep `test.js` in any case. We can consider adding a secondary `test.ts` file, but it worries me. We'd need a commitment from you to keep this up to date with `test.js`. And PRs that add new tests to `test.js`, wouldn't pass until `test.ts` is updated. That's problematic.

vweevers:
> And PRs that add new tests to `test.js`, wouldn't pass until `test.ts` is updated.

Hm, that won't always be the case. More likely, `test.ts` slowly becomes outdated.

zixia:
@vweevers What is the reason for this?
> We have to keep test.js in any case

I'd like to suggest start using the TypeScript for writing the unit tests because it also could do whatever the `.js` could do.

vweevers:
@zixia I can't find it just now but I recall an earlier discussion about keeping the JS. In a nutshell: though we want to make TS a first-class citizen, it does not replace JS. Majority of users and maintainers write in JS.

@ralphtheninja mind weighing in?

ralphtheninja:
> @zixia I can't find it just now but I recall an earlier discussion about keeping the JS. In a nutshell: though we want to make TS a first-class citizen, it does not replace JS. Majority of users and maintainers write in JS.
> 
> @ralphtheninja mind weighing in?

I can only underline this. Bringing in typings to aid people using TypeScript is cool, but if it introduces more complexity and problems where the solution is to rewrite things in TypeScript, then I think we're heading in the wrong way.

Introducing more complexity which causes problems that you solve by more complexity just gives me a bad gut feeling. Call me a dinosaur :smile:

zixia:
So according to the opinions from @vweevers and @ralphtheninja, I believe the only direction now is to try https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript to annotate the JS. 

@Tapppi Good luck!

vweevers:
> I believe the only direction now is to try https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript to annotate the JS.

If that doesn't pan out, would removing all the generics - see https://github.com/Level/abstract-leveldown/pull/195#issuecomment-361270414 and https://github.com/Level/levelup/pull/499#issuecomment-338942816 - also solve this? Or is it unrelated?

sandersn:
Following @zixia's suggestion, you can also add `// @ts-check` in test.js to get Typescript to report all the errors once you have the tests type annotated with jsdoc.

(I'm still trying to understand all the discussion in this thread, after which I may have some other suggestions for getting Typescript to co-operate.)

zixia:
@vweevers i think it's unrelated with the generics.

It's a normal type unmatch.

sandersn:
Here are the jsdocs you need to fix the compile errors:

Line 233, before `var expectedArray`:
`  /** @type {Array<{type: "put", key, value } | { type: "del", key }>} */`

Line 680, before `var op`: 
`  /** @type { {type: 'put', key, value} } */`

This is the bare minimum you need to get `test.js` to compile without errors. It is nicer if you import the type `Batch`. I did this by adding `import { Batch } from './'`or `import * as zelf from './'`. Then you can refer to the type `Batch` in jsdoc comments:

`  /** @type {Array<zelf.Batch>} */`

`  /** @type {zelf.Batch} */`

respectively. Unfortunately, Typescript doesn't support the same type-importing from `var zelf = require('./')` syntax. Fixing that is tracked by bug Microsoft/TypeScript#11825

Tapppi:
@vweevers @zixia

> So according to the opinions from @vweevers and @ralphtheninja, I believe the only direction now is to try https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript to annotate the JS.

That's alright, good to know about the JSDoc support anyway. Also, I agree about converting all of the tests to TS, does not seem like a sensible thing to do in the grander scheme of things, but couldn't hurt to ask 😉 

---

@sandersn


> This is the bare minimum you need to get test.js to compile without errors. It is nicer if you import the type Batch. I did this by adding import * as zelf from './'. Then you can refer to the type Batch in jsdoc comments

Thanks, I think I'll stick with the plain types, to avoid changing style and unused imports. 

> Following @zixia's suggestion, you can also add // @ts-check in test.js to get Typescript to report all the errors once you have the tests type annotated with jsdoc.

Thanks! A few questions if you don't mind:
* Is `// @ts-check` necessary for `ts-node --type-check --no-cache test.js`, or just when using `tsc`?
* do you have any idea what's up with the linux/macos `ts-node` differences? Or is it somehow related to `// @ts-check`.. If we changed to `tsc`, it would be nice to only typecheck, not emit files. Don't know if that's possible
* Do you have any idea why type inference fails for the Batch array? Does typescript never try to infer union constant types? I fail to understand why that would be.

sandersn:
@Tapppi 

1. `// @ts-check` doesn't make a difference for me, actually, because tsconfig has `checkJs: true`. I didn't see that until after posting the above comment. I think ts-node is incorrectly swallowing syntax errors, perhaps, because it reports type errors just fine.
2. I am not sure which linux/macos differences you're referring to. Is that information in the original bug? If you want to use tsc for typechecking only, `tsc --noEmit` (or `"noEmit": true` in tsconfig) will typecheck without emitting.
3. `var x = { a: "hi" }` infers `a: string` unless it has some contextual information that prompts it to expect [immutable] string literals. For a declaration like `var` (or even `const`), only a type annotation would supply context, which is why you have to add those in test.js.

Typescript is pretty conservative, so unless it has evidence that something is immutable, it assumes that it is mutable. If the Typescript project were started today, that might be different, since functional style has become more popular.

Tapppi:
@sandersn

1. Yep, although I was wondering on a general level. Tested it and either `tsconfig.json` `checkJs` or `//@ts-check` IS required for `ts-node` too.
2. @zixia reported that `ts-node` was swallowing the type errors for him [here](https://github.com/Level/abstract-leveldown/pull/204#issuecomment-366930459), but then got them to throw with `tsc` [here](https://github.com/Level/abstract-leveldown/pull/204#issuecomment-366969910)
3. Yeah I get this after rereading typescript mission statements and thinking about it a bit more. Might be an interesting weekend project to try and get TS to not throw for these discriminated unions without changing other behaviour.

vweevers:
General question, going forward. When would we need additional JSDoc comments, and how do we prevent it from blocking PRs? We can't expect JS contributors to add or even know about JSDoc comments.

sandersn:
@vweevers After some discussion with @mhegazy, I think the best approach is to start shipping types from DefinitelyTyped and stop shipping them with the package. This allows contributors not to care about Typescript, and Typescript users can easily submit PRs to Definitely Typed if the types get broken.

This does come with a few downsides:

1. The tests and types on DT need to be updated whenever there's a release. It has the same problem as maintaining test.js AND test.ts. Somebody needs to sign up for this to ensure consistent quality.
2. Removing types from the package is a breaking change; typescript users need to know that they will have to add `@types/abstract-leveldown` to package.json. This should probably wait until a major release.

A few years ago our advice was different because Definitely Typed didn't have nice integration with npm, but now `npm install --save-dev @types/abstract-leveldown` is reasonably simple to add.

Tapppi:
@vweevers 

> General question, going forward. When would we need additional JSDoc comments, and how do we prevent it from blocking PRs?

With current configuration, we'd need them when using typings that TS can't properly infer in tests. Only way to NEVER block PRs because of breaking types is to not typecheck the tests and have more `.ts` tests. Altogether, it's probably best to do as @sandersn suggested and move the typings to `DefinitelyTyped`.

> Removing types from the package is a breaking change; typescript users need to know that they will have to add @types/abstract-leveldown to package.json. This should probably wait until a major release.

Not necessarily. The typings haven't worked for a while in this package, aren't even included in top-level packages (`level-rocksdb`, `levelup`, `level` etc.) since v2 (work ongoing in https://github.com/Level/levelup/pull/499). I was trying to get typings to work for top-level packages from user side and ended up having to define the typings in my own project files.. So there are two questions:

* Do we maintain the typings and `test.js` types or a separate `test.ts` file in this repo until next release, when they are removed in favor of `DefinitelyTyped`?
* Who maintains the `DefinitelyTyped` typings when there are new releases of `*level*`?

This should probably also be reflected across Level packages, since this isn't where most users would be using them.

vweevers:
Before we discuss how to get there (it's not a problem), let's make sure everyone is on board with moving to DefinitelyTyped. Thumbs up? @Level/typescript

vweevers:
I think we can merge this PR in the mean time, but let's answer the above question as soon as possible.

vweevers:
@ralphtheninja are you OK with merging this, and releasing a patch? We can then open a new issue for moving to DefinitelyTyped, and give everybody some time to respond.

ralphtheninja:
251ao4

@vweevers Lets do it :smile:

vweevers:
4.0.3

wookieb:
What if we rewrite levelup to Typescript?

PR #207, opened 2018-02-20
Optional type parameter in index.d.ts:fix syntax

sandersn:
The type `Batch` in index.d.ts has an incorrect `?` following the name of the optional type parameter `V`. The type parameter is already implicitly optional because it has a default type `any`.

This question mark causes Typescript compilation errors. Removing it removes the errors.


sandersn:
Never mind, there's already an open PR for this.

Issue #227, opened 2018-05-23
Consider a factory pattern in test suite


vweevers:
Where `testCommon.factory()` would return a fresh `db`. This allows implementations to add options to their constructor, if they need to. See https://github.com/andrewosh/hyperdown/issues/2 for background.

@andrewosh @ralphtheninja @juliangruber

vweevers:
@andrewosh here's another example of constructor options: https://github.com/Level/level-js. With `levelup`, it can be used like so:

```js
var db = levelup(leveljs('my-db', { prefix: 'my-app-' }))
```

IMO this is much cleaner than `open` options, which would have to be passed through `levelup`. It wouldn't be immediately apparent that the `prefix` option belongs to `level-js`:

```js
var db = levelup(leveljs('my-db'), { prefix: 'my-app-' })
```

@ralphtheninja maybe we should start following the constructor pattern everywhere, including `leveldown`.

ralphtheninja:
> @ralphtheninja maybe we should start following the constructor pattern everywhere, including leveldown.

:+1:

vweevers:
Not yet, there are a lot of angles to consider (and we're currently working on other stuff).

As a temporary solution, you might be able to pass a wrapper to the abstract tests. Along the lines of

```js
require('abstract-leveldown/abstract/get-test').all(wrapper, test)

function wrapper (location) {
  hyperdown.call(this, location, {
    reduce: (a, b) => {
      if (!a) return b
      return a
    }
  }
}

util.inherits(wrapper, hyperdown)
```

andrewosh:
Thanks a bunch! I'll give that a shot

PR #256, opened 2018-07-08
Optional location

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/63

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -6,7 +6,7 @@ var rangeOptions = 'start end gt gte lt lte'.split(' ')
 
 function AbstractLevelDOWN (location) {
   if (!arguments.length || location === undefined) {
-    throw new Error('constructor requires at least a location argument')
+    location = ''
   }
 
   if (typeof location !== 'string') {
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -5,7 +5,7 @@ var hasOwnProperty = Object.prototype.hasOwnProperty
 var rangeOptions = 'start end gt gte lt lte'.split(' ')
 
 function AbstractLevelDOWN (location) {
-  if (!arguments.length || location === undefined) {
+  if (location === undefined) {
     location = ''
   }
 
--- a/test/leveldown-test.js
+++ b/test/leveldown-test.js
@@ -1,10 +1,6 @@
 module.exports.args = function (leveldown, test) {
-  test('test database creation no-arg throws', function (t) {
-    t.throws(
-      leveldown
-      , { name: 'Error', message: 'constructor requires at least a location argument' }
-      , 'no-arg leveldown() throws'
-    )
+  test('test database creation no-arg default location', function (t) {
+    t.equal(leveldown().location, '')
     t.end()
   })
   test('test database creation non-string location throws', function (t) {
--- a/test/leveldown-test.js
+++ b/test/leveldown-test.js
@@ -1,8 +1,4 @@
 module.exports.args = function (leveldown, test) {
-  test('test database creation no-arg default location', function (t) {
-    t.equal(leveldown().location, '')
-    t.end()
-  })
   test('test database creation non-string location throws', function (t) {
     t.throws(
       function () {
@@ -17,6 +13,7 @@ module.exports.args = function (leveldown, test) {
   test('test database open no-arg throws', function (t) {
     var db = leveldown('foo')
     t.ok(db, 'database object returned')
+    t.equal(db.location, 'foo')
     t.ok(typeof db.open === 'function', 'open() function exists')
     t.end()
   })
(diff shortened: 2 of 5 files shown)


abstract-leveldown.js
```diff
@@ -6,7 +6,7 @@ var rangeOptions = 'start end gt gte lt lte'.split(' ')
 
 function AbstractLevelDOWN (location) {
   if (!arguments.length || location === undefined) {
```
ralphtheninja: Not sure why we are checking both these things. Shouldn't checking for `undefined` be enough?

ralphtheninja:
Rebased onto latest master. Needs docs too I guess.

abstract-leveldown.js
vweevers: +1 to removing `!arguments.length`

vweevers:
Can you add a note to UPGRADING.md? We'll need to add type checks to `leveldown` etc

test/leveldown-test.js
```diff
@@ -1,10 +1,6 @@
 module.exports.args = function (leveldown, test) {
-  test('test database creation no-arg throws', function (t) {
-    t.throws(
-      leveldown
-      , { name: 'Error', message: 'constructor requires at least a location argument' }
-      , 'no-arg leveldown() throws'
-    )
+  test('test database creation no-arg default location', function (t) {
+    t.equal(leveldown().location, '')
```
vweevers: hmm, leveldown couldn't pass this test

ralphtheninja: Hmm, maybe this test should be removed then, since I'm guessing `leveldown` messes up on https://github.com/Level/leveldown/blob/master/leveldown.js#L13

vweevers: We could copy the type checks + tests to `leveldown` when we upgrade it.

Or, make `test/leveldown-test` specifically for implementations that do support `location`? Which is already the case, kinda.

ralphtheninja: > Or, make test/leveldown-test specifically for implementations that do support location? Which is already the case, kinda.

:+1:

ralphtheninja:
Browser tests never started, restarting build.

vweevers:
Entertaining two thoughts, because I think we can afford to take bigger steps.

Thought 1. Make `location` optional by checking `arguments.length`. This gives implementations the choice of supporting `location` without them having to do extra type checks. In addition, make the `.location` property optional too?

```js
function AbstractLevelDOWN (location) {
  if (arguments.length > 0) {
    if (typeof location !== 'string') {
      throw new Error('constructor requires a location string argument')
    }

    this.location = location
  }

  this.status = 'new'
}

function MemDOWN () {
   AbstractLevelDOWN.call(this)
}

function LevelDOWN (location) {
  AbstractLevelDOWN.call(this, location)
}

function SQLDown (connection) {
  this._connection = connection
  AbstractLevelDOWN.call(this)
}
```

Thought 2. Remove location altogether, it's not that much code to move to `leveldown`:

```js
function AbstractLevelDOWN () {
  this.status = 'new'
}

function MemDOWN () {
   AbstractLevelDOWN.call(this)
}

function LevelDOWN (location) {
  if (typeof location !== 'string') {
    throw new Error('constructor requires a location string argument')
  }

  this._location = location
  AbstractLevelDOWN.call(this)
}

function SQLDown (connection) {
  this._connection = connection
  AbstractLevelDOWN.call(this)
}
```

An additional benefit of either of these approaches is that we can check `location !== ''`. Are empty locations handled in the c++?

ralphtheninja:
Additionally, I think we might need to change how the abstract tests are invoked as well. If `location` is removed from `abstract-leveldown` and if implementations aren't necessarily using a location, then maybe we shouldn't do e.g.

```js
var db = leveldown(testCommon.location())
```

but rather

```js
var db = factory()
```

So implementations pass in `factory` instead and can handle themselves how the instance should be created, with a location or not.

ralphtheninja:
Aaah we're of course already using a factory, so we shouldn't pass `testCommon.location()` to it.

vweevers:
A `factory` can also support options (#227). Two birds one stone :)

vweevers:
> Aaah we're of course already using a factory

So none of the current tests assume that the function has a prototype?

ralphtheninja:
> So none of the current tests assume that the function has a prototype?

I haven't checked all. With that you mean that we don't have to call `new`?

vweevers:
> With that you mean that we don't have to call new?

And that the tests don't do things like `instanceof fn`

ralphtheninja:
@vweevers If you're cool with it, I'd like to close this PR and start over with _removing_ `location`.

Ps. It seems to me that `test/leveldown-test.js` is getting more and more redundant since it will be only one test left which doesn't really test anything

https://github.com/Level/abstract-leveldown/blob/74d7340dbf85ceae4f69f7bc9ea9d541498c3a8d/test/leveldown-test.js#L21-L26

Issue #200, opened 2018-02-08
AbstractChainedBatch: pass options to _write


vweevers:
Discovered in https://github.com/Level/abstract-leveldown/issues/193#issuecomment-359369932. If `_write` is defined, it does not receive the `options`, unlike `_batch`:

https://github.com/Level/abstract-leveldown/blob/0d942fd45df5c9915269ded6f0f76fdfade11575/abstract-chained-batch.js#L79-L83

Fixing this would be a breaking change. WDYT @ralphtheninja @juliangruber?

juliangruber:
Would it be a breaking change, or rather a long outstanding bug fix? 🤔

vweevers:
It's a breaking change per semver because implementations would have to change their function signature to `_write(options, callback)`.

ralphtheninja:
Hmm I think I've seen an issue on this before, but can't remember where. `leveldown` would not be affected though:

https://github.com/Level/leveldown/blob/c7e2b762de1b99e97d8a86b8d6328da33cec3dc9/chained-batch.js#L26-L28

vweevers:
@ralphtheninja hmm wait, if leveldown's signature is already `_write(options, callback)`.. why does that work?

vweevers:
Ah, the C++ [takes the first argument as a callback](https://github.com/Level/leveldown/blob/c7e2b762de1b99e97d8a86b8d6328da33cec3dc9/src/batch.cc#L134-L135). So it would break leveldown.

PR #257, opened 2018-07-08
Refactor chained batch

vweevers:
Tested against `leveldown` (but it will need some changes, because of https://github.com/Level/abstract-leveldown/issues/200#issuecomment-364128603).

Left to do:

- [x] Update API docs
- [x] ~~Add to `UPGRADING.md`~~ => #262

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ Free up underlying resources. This method is guaranteed to only be called once. 
 
 ### `chainedBatch = AbstractChainedBatch(db)`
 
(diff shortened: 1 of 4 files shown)


vweevers:
@ralphtheninja as you're working on the upgrade guide (#262), I'll wait with adding the changes from this PR.

vweevers:
@ralphtheninja can you review this one? I don't want to keep rebasing ;)

PR #258, opened 2018-07-10
Remove location

### LINKED ISSUE #63: make location optional

ralphtheninja:
Closes #63 

Also see discussion from #256

### DIFF
--- a/test/open-test.js
+++ b/test/open-test.js
@@ -84,7 +84,6 @@ module.exports.openAdvanced = function (factory, test) {
   test('test database open errorIfExists:true', function (t) {
     var db = factory()
 
-    // make a valid database first, then close and dispose
     db.open({}, function (err) {
       t.error(err)
       db.close(function (err) {
@@ -92,7 +91,6 @@ module.exports.openAdvanced = function (factory, test) {
 
         var async = false
 
-        // open again with 'errorIfExists'
         db.open({ createIfMissing: false, errorIfExists: true }, function (err) {
           t.ok(err, 'error')
           t.ok(/exists/.test(err.message), 'error is about already existing')


test/leveldown-test.js
```diff
@@ -1,23 +1,4 @@
...
-      function () {
-        leveldown({})
-      }
-      , { name: 'Error', message: 'constructor requires a location string argument' }
-      , 'non-string location leveldown() throws'
-    )
-    t.end()
-  })
-
   test('test database open no-arg throws', function (t) {
```
ralphtheninja: What to do with this file now? It only contains one test which doesn't do anything. I think we have two options here.

1. We remove it completely
2. We rewrite it to test other things that might be related to inheritance, such as checking for the complete api and not only `.open`
3. Keep it as it is

If we go with 2 or three, the current test should be renamed because it doesn't have anything to do with `throw`

ralphtheninja: I personally would go for 1.

ralphtheninja:
Restarting browser tests a _second_ time.

vweevers:
What about the abstract tests?

https://github.com/Level/abstract-leveldown/blob/74d7340dbf85ceae4f69f7bc9ea9d541498c3a8d/test/batch-test.js#L8

etc

ralphtheninja:
@vweevers I thought I'd do a second PR for that.

ralphtheninja:
Can fix in this PR as well if you want.

ralphtheninja:
Found this test which assumes location based implementations

https://github.com/Level/abstract-leveldown/blob/74d7340dbf85ceae4f69f7bc9ea9d541498c3a8d/test/open-test.js#L84-L109

Will comment out this for now.

test/open-test.js
```diff
@@ -81,43 +81,45 @@ module.exports.openAdvanced = function (leveldown, test, testCommon) {
...
-          t.ok(/exists/.test(err.message), 'error is about already existing')
-          t.ok(async, 'callback is asynchronous')
-          t.end()
-        })
-
-        async = true
-      })
-    })
-  })
+  // TODO this assumes location based backends, move out to respective
```
ralphtheninja: This is the only test that assumes the (previous) `leveldown` parameter to accept a `location` argument (so it can try and open on the same location). We should either remove this and have location based implementations use it, or move it to a location based file (e.g. `test/leveldown-test.js`).

However, it uses `errorIfExists` property which afaik is related to `leveldown`, so maybe we should just move it to `LevelDB` based implementations.

ralphtheninja:
@vweevers I could use some feedback on the `TODO` comments specifically.

ralphtheninja:
Left to do:

- [ ] some notes in `UPGRADING.md` on api changes
- [ ] address TODO comments

test/leveldown-test.js
vweevers: +1 to removing

test/open-test.js
vweevers: Hm, why does this test need two `db` instances? If you open a db, close it, then reopen with `errorIfExists`, it should error too, right?

(Removing `createIfMissing` and `errorIfExists` from `abstract-leveldown` is probably a discussion for later)

README.md
```diff
@@ -45,9 +45,10 @@ Let's implement a simplistic in-memory [`leveldown`][leveldown] replacement:
 var AbstractLevelDOWN = require('abstract-leveldown').AbstractLevelDOWN
 var util = require('util')
 
-// Constructor, passes location to the AbstractLevelDOWN constructor
+// Constructor
 function FakeLevelDOWN (location) {
-  AbstractLevelDOWN.call(this, location)
+  this.location = location
```
vweevers: Might as well remove this too

test/common.js
```diff
@@ -1,9 +1,8 @@
-var collectEntries = require('level-concat-iterator')
 var tempy = require('tempy')
 
 module.exports = {
+  // TODO remove?
```
vweevers: Removing SGTM. Implementations only have to call `tempy.directory` in their factory.

vweevers:
Would it make sense for `factory` to be part of `testCommon`?

test/common.js
ralphtheninja: Yep! Also means we can merge `common.js` and `common-browser.js` into one file.

vweevers: Yay!

vweevers: Oh but we should document somewhere that each `factory` call should return a fresh db

vweevers: Or add a test for this behavior :)

vweevers:
> Maybe merge those in another PR?

👍

vweevers:
I love where this is going!

ralphtheninja:
> I love where this is going!

Haha same here. Things just seemed to fall in place a lot when location was no longer something to consider.

test/open-test.js
```diff
@@ -80,35 +80,6 @@ module.exports.openAdvanced = function (factory, test) {
 
     async = true
   })
-
-  // TODO this assumes location based backends, move out to respective
```
vweevers: I'm not sure about this, see https://github.com/Level/abstract-leveldown/pull/258#discussion_r201442144

test/open-test.js
ralphtheninja: I'm in a rush right now. Will get back to this in a bit.

ralphtheninja: @vweevers Reverted the commit and brought back the test using a single db instance.

ralphtheninja:
I added the tag `pre-remove-location` to current master. Creating a `abstract-leveldown-v6` branch in `leveldown` to try out the new changes to see if we are running into something unforeseen. Keeping WIP status of this PR for a while.

ralphtheninja:
Rebased onto latest master (without node 9 so faster builds yay!)

test/common.js
```diff
@@ -1,9 +1,4 @@
-var collectEntries = require('level-concat-iterator')
-var tempy = require('tempy')
-
 module.exports = {
```
ralphtheninja: Maybe we could export a function here instead? E.g.

```js
module.exports = function (opts) {
  return xtend({
    setUp: function (t) { t.end() },
    tearDown: function (t) { t.end() }
  }, opts)
}
```

ralphtheninja:
@vweevers Sorry to be a pain. Good to squash?

vweevers:
Can you give me until tonight (~6 hours from now)?

ralphtheninja:
> Can you give me until tonight (~6 hours from now)?

For sure!

test/open-test.js
```diff
@@ -82,20 +82,17 @@ module.exports.openAdvanced = function (leveldown, test, testCommon) {
   })
 
   test('test database open errorIfExists:true', function (t) {
-    var location = testCommon.location()
-    var db = leveldown(location)
+    var db = factory()
 
     // make a valid database first, then close and dispose
```
vweevers: Leftover comment: " and dispose"

vweevers:
@ralphtheninja Can you open issues for:

- Document that each factory call must return a fresh db
- Test that each factory call returns a fresh db (open two dbs, write to first, check that second is empty)
- Consider exporting testCommon as function
- Update upgrade guide (also for `seek`)

test/open-test.js
ralphtheninja: Removing the comments altogether.

PR #268, opened 2018-07-11
Merge factory function into testCommon parameter

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/227

### DIFF
--- a/test.js
+++ b/test.js
@@ -8,7 +8,7 @@ var AbstractLevelDOWN = require('./').AbstractLevelDOWN
 var AbstractIterator = require('./').AbstractIterator
 var AbstractChainedBatch = require('./').AbstractChainedBatch
 
-function factory () {
+var factory = testCommon.factory = function () {
   return new AbstractLevelDOWN()
 }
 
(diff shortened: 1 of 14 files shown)


test.js
```diff
@@ -8,53 +8,53 @@ var AbstractLevelDOWN = require('./').AbstractLevelDOWN
...
-require('./test/iterator-test').setUp(factory, test, testCommon)
-require('./test/iterator-test').args(test)
-require('./test/iterator-test').sequence(test)
+require('./test/iterator-test').setUp(test, testCommon)
+require('./test/iterator-test').args(test, testCommon)
+require('./test/iterator-test').sequence(test, testCommon)
 require('./test/iterator-test').tearDown(test, testCommon)
 
-require('./test/iterator-range-test').setUp(factory, test, testCommon, [])
+require('./test/iterator-range-test').setUp(test, testCommon, [])
```
ralphtheninja: Now the api is _almost_ identical for _all_ test functions, except for two in `test/iterator-range-test.js` which takes an additional `data` parameter. Which bugs me a bit :smile:

test/batch-test.js
```diff
@@ -266,11 +266,11 @@ module.exports.tearDown = function (test, testCommon) {
...
 
-module.exports.all = function (factory, test, testCommon) {
+module.exports.all = function (test, testCommon) {
   testCommon = testCommon || require('./common')
-  module.exports.setUp(factory, test, testCommon)
-  module.exports.args(test)
-  module.exports.batch(test)
-  module.exports.atomic(test)
+  module.exports.setUp(test, testCommon)
+  module.exports.args(test, testCommon)
```
ralphtheninja: I decided to tack on `testCommon` everywhere, even if it's not used in some tests. Just to make it more symmetric.

vweevers: Doesn't standard complain about that?

ralphtheninja: I thought so as well.

test.js
vweevers: We could make that internal by only exporting `all`, because this file only has `setUp`, `range` and `tearDown` - that's one test. We only have to export "subtests" if there are multiple, and if they should be skippable.

I don't know if now is the right time, but we can also make an effort to split tests files in such a way that you'd only exclude files, not subtests. Then each file only needs one export.

vweevers: => #270

Issue #264, opened 2018-07-11
Remove default testCommon parameter


ralphtheninja:
Previously we often just used `testCommon.js` from `abstract-leveldown` and passed it in to abstract tests and it made sense at the time to make it a default parameter, e.g.

```js
module.exports.all = function (test, testCommon) {
  testCommon = testCommon || require('./common')
  // ..
}
```

This doesn't really make much sense now since all implementations are forced to pass in their own `testCommon` due to factory function etc.

ralphtheninja:
Done.

Issue #259, opened 2018-07-11
Document that each factory call must return a fresh db


ralphtheninja:
See discussion in https://github.com/Level/abstract-leveldown/pull/258

Maybe also update the docs with an example of how to run the abstract tests.

vweevers:
Repurposed this issue to be about general documentation of the test suite.

PR #271, opened 2018-07-13
Export test suite as single function

### LINKED ISSUE #157: Export simple interface to entire abstract suite

vweevers:
Do not merge, ~~I'm just exploring~~. This will make it much easier to update implementations, because they only have to call one function. @ralphtheninja WDYT?

Closes #157.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -15,21 +15,8 @@ var common = require('./common')
  * ```
  */
 module.exports = function (options) {
-  var test = options.test
-  var factory = options.factory
-
-  // TODO: consider renaming to "setup" and "teardown" (lowercase)
-  var setUp = options.setUp || common.setUp
-  var tearDown = options.tearDown || common.tearDown
-
-  if (!factory) throw new TypeError('factory function is required')
-  if (!test) throw new TypeError('test function is required')
-
-  var testCommon = {
-    factory: factory,
-    setUp: setUp,
-    tearDown: tearDown
-  }
+  var testCommon = common(options)
+  var test = testCommon.test
 
   require('./leveldown-test').args(test, testCommon)
   require('./open-test').all(test, testCommon)
--- a/test/index.js
+++ b/test/index.js
@@ -2,19 +2,16 @@ module.exports = function (test, testCommon, options) {
   options = options || {}
 
   require('./leveldown-test').args(test, testCommon)
-
-  // TODO: move openAdvanced tests to separate files so they can be skipped.
   require('./open-test').all(test, testCommon)
-
-  // if (options.createIfMissing !== false) {
-  //   require('./open-create-if-missing-test').all(test, testCommon)
-  // }
-  //
-  // if (options.errorIfExists !== false) {
-  //   require('./open-error-if-exists-test').all(test, testCommon)
-  // }
-
   require('./close-test').close(test, testCommon)
+
+  if (options.createIfMissing !== false) {
+    require('./open-create-if-missing-test').all(test, testCommon)
+  }
+
+  if (options.errorIfExists !== false) {
+    require('./open-error-if-exists-test').all(test, testCommon)
+  }
 
   require('./put-test').all(test, testCommon)
   require('./get-test').all(test, testCommon)
--- a/test/index.js
+++ b/test/index.js
@@ -1,0 +1,35 @@
+module.exports = function (test, testCommon, options) {
+  options = options || {}
+
+  require('./leveldown-test').args(test, testCommon)
+
+  // TODO: move openAdvanced tests to separate files so they can be skipped.
+  require('./open-test').all(test, testCommon)
+
+  // if (options.createIfMissing !== false) {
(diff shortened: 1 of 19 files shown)


ralphtheninja:
Hehe this goes in the direction of how `level-packager` used to be. But I actually like it because there are so many tests it's helpful.

vweevers:
What do you like better, before or after 45dd8a8? On the one hand a single options object is cleaner but I don't like that it has required "options".

ralphtheninja:
> What do you like better, before or after 45dd8a8? On the one hand a single options object is cleaner but I don't like that it has required "options".

After. Mostly because implementations don't need to explicitly require `test/common`, modify it and then pass it back in.

vweevers:
> Mostly because implementations don't need to explicitly require test/common, modify it and then pass it back in.

Roger. In fact, we can remove `test/common` altogether..

ralphtheninja:
> Roger. In fact, we can remove test/common altogether..

Aye, it could just be part of `test/index.js`.

vweevers:
I want a gif for how everything is falling into place

ralphtheninja:
falling-into-place

vweevers:
... not what I was expecting 🤣

ralphtheninja:
> Hmm, test.js does still need testCommon.

We could also consider duplicating the code, noop `setUp()` and `tearDown()` methods isn't a lot of code.

vweevers:
I opted to move the options logic to `test/common` as a function, will push shortly.

ralphtheninja:
@vweevers Will you close https://github.com/Level/abstract-leveldown/issues/264 in this PR as well?

vweevers:
> Will you close #264 in this PR as well?

Sure!

test/index.js
```diff
@@ -0,0 +1,49 @@
+var common = require('./common')
+
+/**
+ * Example usage:
+ *
+ * ```
+ * require('abstract-leveldown/test')({
```
vweevers: Except, this will resolve to `test.js`

vweevers: @ralphtheninja I just like `require('abstract-leveldown/test')` too much, so what's a good new name for `test.js`?

ralphtheninja: We can rename local `test.js` to something else.

vweevers: It should still match nyc/instanbul default ignore patterns. Maybe `test/self.js`?

ralphtheninja: Naming is always the hardest :smile: How about `test-local.js`?

vweevers: Should match both:

- `['**/node_modules/**', '**/bower_components/**', '**/test/**', '**/tests/**', '**/*.json']` (browserify-instanbul)
- `['coverage/**', 'test/**', 'test{,-*}.js', '**/*.test.js', '**/__tests__/**', '**/node_modules/**']` (nyc)

So the file must be in `test/`.

vweevers:
I'll continue after a short break :) We're on a roll!

vweevers:
> Maybe a short thing should go into README as well, a simple example of how to call the test function.

Can do that as part of #259

Issue #270, opened 2018-07-13
Remove exports from iterator-range-test


vweevers:
See https://github.com/Level/abstract-leveldown/pull/268#discussion_r201867273.

ralphtheninja:
Hmm `setUp()` and `tearDown()` are still called separately from internal tests so we can't export a single method `.all()`.

PR #275, opened 2018-07-13
Lift out data to file scope to get rid of extra parameter

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/270

I solved this differently, since `test/self.js` is calling `setUp()` and `tearDown()` we can't export a single `.all()` method.

Upgrade guide needs updating, because now _all_ test methods have the same api.

### DIFF
--- a/test/iterator-range-test.js
+++ b/test/iterator-range-test.js
@@ -2,7 +2,21 @@ var collectEntries = require('level-concat-iterator')
 
 var db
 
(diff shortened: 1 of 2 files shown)


vweevers:
> Upgrade guide needs updating, because now all test methods have the same api.

I figure we can just remove that section, because `abstract-leveldown/test` is now the preferred entry point. The internals no longer matter for consumers.

Issue #222, opened 2018-05-17
make _setupIteratorOptions part of the official api


ralphtheninja:
See https://github.com/Level/encoding-down/pull/45 for details.

Add it to the docs and explain what it does.

ralphtheninja:
See discussion in https://github.com/Level/abstract-leveldown/issues/120

PR #277, opened 2018-07-13
Serialization

### LINKED ISSUE #120: drop storing null and undefined values
EDIT: Poll locked.

screenshot from 2018-07-08 12-58-34

Also see discussion here https://github.com/Level/abstract-leveldown/pull/119

### LINKED ISSUE #130: range options should be serialized
See https://github.com/Level/abstract-leveldown/issues/121#issuecomment-333484988.

@ralphtheninja @juliangruber should we include this in `levelup@2`? If so I can try to tackle this today/tomorrow.

### LINKED ISSUE #230: Change error message for empty array key
https://github.com/Level/abstract-leveldown/blob/a883f3bdb93d94dc26207893f6e9c782ab010644/abstract/put-get-del-test.js#L107

Because implementations like `level-js` support array keys, and it's weird that the error would say "empty String".

vweevers:
Closes #120, closes #130, closes #230. Checklist adapted from https://github.com/Level/abstract-leveldown/issues/120#issuecomment-390171073:

- [x] Don't touch nullish range options in `abstract#_setupIteratorOptions` (no need for #222)
- [x] Don't touch empty range options in `abstract#_setupIteratorOptions`. ~~TBD~~
- [x] Change `_serializeKey` to an identity function
- [x] Change `_serializeValue` to an identity function
- [x] ~~Reject nullish keys *before* calling `_serializeKey`.~~ (Already the case). The benefit is that `_serializeKey` can be used on range options with nullish support.
- [x] Reject nullish values *before* calling `_serializeValue`
- [x] ~~Reject nullish and empty keys *after* calling `_serializeKey`? TBD~~
- [x] ~~Reject nullish values *after* calling `_serializeValue`? TBD~~
- [x] Pass range options through `_serializeKey` (#130)
- [x] Remove boolean and NaN key tests

These are very tricky changes, will need canary testing in at least `leveldown`.

Externally (just mental notes):

- Add `leveldown#_serializeKey` and `#_serializeValue` (same logic) to convert nullish to an empty string, Buffer as-is, everything else to a string.

### DIFF
--- a/test/chained-batch-test.js
+++ b/test/chained-batch-test.js
@@ -95,6 +95,19 @@ exports.args = function (test, testCommon) {
     t.end()
   })
 
+  test('test batch#put() with null or undefined `value`', function (t) {
+    var illegalValues = [null, undefined]
+    t.plan(illegalValues.length)
+
+    illegalValues.forEach(function (value) {
+      try {
+        db.batch().put('key', value)
+      } catch (err) {
+        t.equal(err.message, 'value cannot be `null` or `undefined`', 'correct error message')
+      }
+    })
+  })
+
   test('test batch#del() with missing `key`', function (t) {
     try {
       db.batch().del()
--- a/test/chained-batch-test.js
+++ b/test/chained-batch-test.js
@@ -106,15 +106,17 @@ exports.args = function (test, testCommon) {
     t.end()
   })
 
-  test('test batch#del() with null `key`', function (t) {
-    try {
-      db.batch().del(null)
-    } catch (err) {
-      t.equal(err.message, 'key cannot be `null` or `undefined`', 'correct error message')
-      return t.end()
-    }
-    t.fail('should have thrown')
-    t.end()
+  test('test batch#del() with null or undefined `key`', function (t) {
+    var illegalKeys = [null, undefined]
+    t.plan(illegalKeys.length)
+
+    illegalKeys.forEach(function (key) {
+      try {
+        db.batch().del(key)
+      } catch (err) {
+        t.equal(err.message, 'key cannot be `null` or `undefined`', 'correct error message')
+      }
+    })
   })
 
   test('test batch#clear() doesn\'t throw', function (t) {
--- a/test/chained-batch-test.js
+++ b/test/chained-batch-test.js
@@ -117,17 +117,6 @@ exports.args = function (test, testCommon) {
     t.end()
   })
 
-  test('test batch#del() with null `key`', function (t) {
-    try {
-      db.batch().del(null)
-    } catch (err) {
-      t.equal(err.message, 'key cannot be `null` or `undefined`', 'correct error message')
-      return t.end()
-    }
-    t.fail('should have thrown')
-    t.end()
-  })
-
   test('test batch#clear() doesn\'t throw', function (t) {
     db.batch().clear()
     t.end()
--- a/test/chained-batch-test.js
+++ b/test/chained-batch-test.js
@@ -39,13 +39,27 @@ exports.args = function (test, testCommon) {
   })
 
   test('test batch#put() with missing `value`', function (t) {
-    db.batch().put('foo1')
-    t.end()
+    t.plan(1)
+
+    try {
+      db.batch().put('foo1')
+    } catch (err) {
+      t.is(err.message, 'value cannot be `null` or `undefined`', 'correct error message')
+    }
   })
 
-  test('test batch#put() with null `value`', function (t) {
-    db.batch().put('foo1', null)
-    t.end()
+  test('test batch#put() with null or undefined `value`', function (t) {
+    var illegalValues = [null, undefined]
+
+    t.plan(illegalValues.length)
+
+    illegalValues.forEach(function (value) {
+      try {
+        db.batch().put('foo1', value)
+      } catch (err) {
+        t.is(err.message, 'value cannot be `null` or `undefined`', 'correct error message')
+      }
+    })
   })
 
   test('test batch#put() with missing `key`', function (t) {
--- a/test/chained-batch-test.js
+++ b/test/chained-batch-test.js
@@ -48,20 +48,6 @@ exports.args = function (test, testCommon) {
     }
   })
 
-  test('test batch#put() with null or undefined `value`', function (t) {
-    var illegalValues = [null, undefined]
-
-    t.plan(illegalValues.length)
-
-    illegalValues.forEach(function (value) {
-      try {
-        db.batch().put('foo1', value)
-      } catch (err) {
-        t.is(err.message, 'value cannot be `null` or `undefined`', 'correct error message')
-      }
-    })
-  })
-
   test('test batch#put() with missing `key`', function (t) {
     try {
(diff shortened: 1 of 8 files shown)


vweevers:
Should also allow empty range options (zero-length string or zero-length Buffer) in `_setupIteratorOptions`.

As @dominictarr wrote in https://github.com/Level/encoding-down/issues/42:

> \[abstract-leveldown now\] strips out nullish ranges, which is reasonable for leveldown, but was supported in levelup, and null, undefined and '' are significant values in some encodings (bytewise, charwise).

Empty range options should be OK for leveldown, because IIRC it ignores such options in the C++.

vweevers:
Problem: in `encoding-down`, range options must be serialized *after* encoding them.

vweevers:
> Problem: in encoding-down, range options must be serialized after encoding them.

No :) The order doesn't matter in `encoding-down`, as its `_serializeKey` is an identify function.

vweevers:
@ralphtheninja WDYT about:

- Reject nullish and empty keys *after* calling `_serializeKey`?
- Reject nullish values *after* calling `_serializeValue`?

Note that before this PR (specifically, dec6278), keys that would *likely* serialize to an empty string were rejected:

https://github.com/Level/abstract-leveldown/blob/d7411bbef7c7b8bd5f144db24b35a6cbdcd918f6/abstract-leveldown.js#L247-L249

And this `String(obj)` call was kinda wasteful.

vweevers:
I'm leaning towards a yes, because we gotta make sure we're not passing empty string keys to leveldown, which segfaults on empty slices, or at least it used to. @ralphtheninja is that still the case, or does the C++ guard against that nowadays?

ralphtheninja:
> is that still the case, or does the C++ guard against that nowadays?

Hmm, I would have to check.

vweevers:
Might be moot. I'm thinking if leveldown doesn't support empty keys, other implementations shouldn't either for consistency. So it's something that `abstract-leveldown` should guard against.

ralphtheninja:
```js
var leveldown = require('.')                         
var db = leveldown('./db')                           
db.open(function (err) {                             
  db.put('', 'VALUE', function (err) {               
    console.log('after put', err)                    
    db.get('', function (err, value) {               
      console.log('after get', err, value.toString())
    })                                               
  })                                                 
})                                                   
```

If I comment out `_checkKey` in `abstract-leveldown` I get:

```
after put undefined
after get null VALUE
```

vweevers:
@ralphtheninja huh, I did not expect that!

vweevers:
We have three options, in pseudocode:

- `serialize(check(key))` - this doesn't fully guard against empty keys, might be acceptable though
- `check(serialize(key))` - this can't block types that we want to be illegal across the board
- `check(serialize(check(key)))` - overkill?

Oh, how I hate this.

vweevers:
I made up my mind. I want to keep the current logic - `serialize(check(key))` - because serialization is meant to convert a key to a type supported by the underlying storage. That can be *any* type, we shouldn't care. We check user input, yes, but once we pass data to be serialized, it's out of our hands.

What matters more, is how the data *comes out* of the storage - i.e. it shouldn't be nullish because of preexisting significance in iterators and streams. We don't have a runtime guard against that atm, or a good mechanism in place to support such a guard. Adding more input checks certainly doesn't help, and we can't write abstract tests for it because we can't make assumptions about serialization or type support. So... that's something to attack later (or not).

For now, rejecting nullish as *input* should be enough of a signal to consumers and implementors. In addition, I'd like to keep rejecting empty keys, simply because that's how it's always been.

The only gap left is that before this PR, `abstract-leveldown` rejected `String(key) === ''`. This is a real edge case though, because all JS types for which `String(key) === ''` are already rejected: strings, arrays and buffers. The only type that can pass through is an object with a custom `toString()` that returns `''`.

vweevers:
Tested against `leveldown` with one addition:

```js
LevelDOWN.prototype._serializeKey =
LevelDOWN.prototype._serializeValue = function (datum) {
  return Buffer.isBuffer(datum) ? datum : String(datum)
}
```

vweevers:
After this I think we should add `next` branches to `memdown` and `level-js`.

test/chained-batch-test.js
```diff
@@ -81,6 +95,19 @@ exports.args = function (test, testCommon) {
     t.end()
   })
 
+  test('test batch#put() with null or undefined `value`', function (t) {
```
vweevers: Duplicate test :)

test/iterator-range-test.js
```diff
@@ -337,29 +337,13 @@ exports.range = function (test, testCommon) {
...
     end: ''
   }, data)
-
-  rangeTest('test iterator with lte as null', {
-    lte: null
-  }, data)
-
-  rangeTest('test iterator with end as null - legacy', {
-    end: null
-  }, data)
```
ralphtheninja: @vweevers Why were these tests removed?

test/put-get-del-test.js
```diff
@@ -104,20 +104,23 @@ exports.errorKeys = function (test, testCommon) {
   makeErrorKeyTest(test, 'undefined key', undefined, /key cannot be `null` or `undefined`/)
   makeErrorKeyTest(test, 'empty String key', '', /key cannot be an empty String/)
   makeErrorKeyTest(test, 'empty Buffer key', Buffer.alloc(0), /key cannot be an empty \w*Buffer/)
-  makeErrorKeyTest(test, 'empty Array key', [], /key cannot be an empty String/)
+  makeErrorKeyTest(test, 'empty Array key', [], /key cannot be an empty Array/)
+}
+
+exports.errorValues = function (test, testCommon) {
+  makePutErrorTest(test, 'null value', 'key', null, /value cannot be `null` or `undefined`/)
+  makePutErrorTest(test, 'undefined value', 'key', undefined, /value cannot be `null` or `undefined`/)
```
ralphtheninja: Aaaaand it's back! :)

test/iterator-range-test.js
vweevers: Because we can't really say what `null` should mean. In `bytewise` for example, `null` is a lower bound, so `end: null` wouldn't match anything.

test/put-get-del-test.js
vweevers: This is not related to the removal of the rangeTest

ralphtheninja: I know. Just funny that `errorValues` came back with an actual usage (since it was empty and removed).

vweevers: I misunderstood because your two comments followed each other :)

Jeez, I didn't even realize that `errorValues` is the one we removed 🤣

vweevers:
> After this I think we should add next branches to memdown and level-js.

And `encoding-down` (where we should be able to remove `_setupIteratorOptions`).

vweevers:
done-downsized

ralphtheninja:
Was nice to finally get this merged!

ralphtheninja:
HAHAHAHA there we have it. A perfect gif. I think we should use it in the release notes!

Issue #260, opened 2018-07-11
Test that each factory call returns a fresh db


ralphtheninja:
(open two dbs, write to first, check that second is empty)

See discussion in #258

PR #282, opened 2018-07-14
Test that each factory returns a unique db

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/260

### DIFF
--- a/test/factory-test.js
+++ b/test/factory-test.js
@@ -1,0 +1,33 @@
+var concat = require('level-concat-iterator')
+
+module.exports = function (test, testCommon) {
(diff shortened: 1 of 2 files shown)


vweevers:
Shouldn't each test call setup and teardown?

ralphtheninja:
> Shouldn't each test call setup and teardown?

Guess so. Not all test are doing that still.

vweevers:
Or hm, is any implementation actually using setup/teardown? 🤔

ralphtheninja:
> Or hm, is any implementation actually using setup/teardown?

We would have to investigate this further I think.

PR #288, opened 2018-07-14
Add checks for null options (since typeof null === "object")

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/283

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -12,13 +12,13 @@ AbstractLevelDOWN.prototype.open = function (options, callback) {
   var self = this
   var oldStatus = this.status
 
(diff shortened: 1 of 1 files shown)


abstract-leveldown.js
```diff
@@ -142,7 +142,7 @@ AbstractLevelDOWN.prototype.batch = function (array, options, callback) {
     return process.nextTick(callback, new Error('batch(array) requires an array argument'))
   }
 
-  if (!options || typeof options !== 'object') { options = {} }
+  if (typeof options !== 'object' || options === null) options = {}
```
ralphtheninja: Changed this for consistency

PR #291, opened 2018-07-14
Update changelog for v6

ralphtheninja:
I ran through this pretty quickly so it's most likely a bit rough. PTAL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,6 @@
 * Remove copyright headers from code (@ralphtheninja)
 * Remove node 9 from Travis (@ralphtheninja)
 * Remove `location` (#258) (@ralphtheninja)
-* Remove deprecation warning on moved snapshot test (@ralphtheninja)
 
 ### Fixed
 * Make sure all `t.throw` tests check error messages correctly (@ralphtheninja)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,37 @@
 
 ## [Unreleased]
 
+### Changed
+* Upgrade `airtap` devDependency from `0.0.5` to `0.1.0` (@vweevers, @ralphtheninja, @greenkeeper)
+* Upgrade `sinon` devDependency from `^5.0.0` to `^6.0.0` (@ralphtheninja)
+* Tweak copyright years for less maintenance (@ralphtheninja)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,17 +2,48 @@
...
 
-### Added
-* Add node 10 to Travis (@ralphtheninja)
-* Add `airtap` for browser tests (@vweevers)
+### Removed
+* Remove `contributors` from `package.json` (@ralphtheninja)
+* Remove copyright headers from code (@ralphtheninja)
+* Remove node 9 from Travis (@ralphtheninja)
+* Remove `location` (#258) (@ralphtheninja)
+* Remove deprecation warning on moved snapshot test (@ralphtheninja)
```
vweevers: Can leave this out; there's no change between v5 and v6

CHANGELOG.md
```diff
@@ -2,17 +2,48 @@
...
-## [5.0.0] - 2018-05-22
+### Changed
+* Upgrade `airtap` devDependency from `0.0.5` to `0.1.0` (@vweevers, @ralphtheninja, @greenkeeper)
+* Upgrade `sinon` devDependency from `^5.0.0` to `^6.0.0` (@ralphtheninja)
+* Tweak copyright years for less maintenance (@ralphtheninja)
+* Isolate snapshot test (#239) (@vweevers)
+* Use `level-concat-iterator` instead of `collectEntries` (@ralphtheninja)
+* Rename `abstract/` to `test/` (@vweevers)
+* Write API docs (#251) (@vweevers)
+* Rewrite `testCommon` functionality and force implementations to implement `testCommon.factory` (@ralphtheninja)
```
vweevers: Hmm I now realize the upgrade guide says `testCommon` is gone but that's not really true.

vweevers: As you demonstrated in leveldown, there's still a use for testCommon by itself.

vweevers: Idea: add the skip options (`snapshots` etc) to `test/common.js` too, so that "testCommon" is the same as the "options" passed to `test/index.js`. Then we can refer to both as "testCommon" in the upgrade guide.

vweevers: => #292

vweevers:
LGTM

PR #293, opened 2018-07-15
Merge testCommon and test options

### LINKED ISSUE #292: Merge testCommon and test options
- [ ] Add the skip options to `test/common.js`, with defaults
- [ ] Remove the lowercase aliases of `setUp` and `tearDown` (I still prefer lowercase but we should either rename everything or leave as-is):

https://github.com/Level/abstract-leveldown/blob/b4e753056d3a861be556b6f44e792d4cac6501ea/test/common.js#L16-L17

vweevers:
Closes #292 and includes some more tests in `test/self`.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -10,11 +10,11 @@ function suite (options) {
   require('./open-test').all(test, testCommon)
   require('./close-test').all(test, testCommon)
 
-  if (options.createIfMissing !== false) {
+  if (testCommon.createIfMissing) {
     require('./open-create-if-missing-test').all(test, testCommon)
   }
 
-  if (options.errorIfExists !== false) {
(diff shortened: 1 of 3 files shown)


test/index.js
```diff
@@ -24,11 +10,11 @@ module.exports = function (options) {
   require('./open-test').all(test, testCommon)
   require('./close-test').all(test, testCommon)
 
-  if (options.createIfMissing !== false) {
+  if (testCommon.createIfMissing) {
```
ralphtheninja: Aaah _now_ I see what you mean.

vweevers: Yea, it's hard to explain changes to things we didn't have words for yet ;)

Issue #285, opened 2018-07-14
.sync() not called from .all() in put-test.js


ralphtheninja:
https://github.com/Level/abstract-leveldown/blob/48484ce12e8e59c0665b50b4bc085809bd662cf4/test/put-test.js#L97-L117

Does this has any implications? Should it be moved to `leveldown`?

vweevers:
> Should it be moved to leveldown?

I vote yes.

But please wait with touching the tests, I'm currently updating them for #277.

vweevers:
Ah, it's fine, I'll rebase. I don't want to rush #277, but it also shouldn't block other tasks

vweevers:
While we're at it, I don't understand what this `sync` test actually tests. It does two tests. The first is just a `put()` with a `sync` option, but the result would be the same if you leave out `sync`. Might be better to test that the `sync` option is passed on to `_put()`. The second test seems to assume that `sync` means synchronous, but it isn't (in Node.js).

ralphtheninja:
I don't either. Doesn't really have anything to do with the `sync` property.

PR #300, opened 2018-07-15
Remove sync test

ralphtheninja:
Closes https://github.com/Level/abstract-leveldown/issues/285


vweevers:
@ralphtheninja reading https://github.com/Level/leveldown/issues/157 and the PR that introduced these `sync` tests (https://github.com/Level/abstract-leveldown/pull/58), I now understand they were about triggering segfaults. I just tested it on latest leveldown, still segfaults. So I'll add the test to leveldown instead, but skip it, and add a comment with some background.

PR #309, opened 2018-10-19
Make db property mandatory and public on iterator and chained batch

### LINKED ISSUE #255: Make db argument of iterator and chained batch required
So that iterators and chained batches will always have a `db` property (or `_db`, normalizing that is a second question). This prevents garbage collection of `db`.

https://github.com/Level/abstract-leveldown/blob/7b079c5510f0ae354ae9503849ac7d9a0aa01551/abstract-iterator.js#L1-L2

https://github.com/Level/abstract-leveldown/blob/7b079c5510f0ae354ae9503849ac7d9a0aa01551/abstract-chained-batch.js#L1-L2

---

- [x] iterator
  - [x] make it required
  - [x] rename to `.db` (already the case)
  - [x] document
- [x] chained batch:
  - [x] make it required (#257)
  - [x] rename to `.db`
  - [x] document

vweevers:
Closes #255.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -153,6 +153,10 @@ suite({
 
 Please see the [README](README.md) for a list of options. Note that some of these have replaced `process.browser` checks.
 
+### Iterator must have a `db` reference
+
+The `db` argument of the `AbstractIterator` constructor became mandatory, as well a public `db` property on the instance. Its existence is not new; the test suite now asserts that your implementation also has it.
+
 ### Seeking became part of official API
 
 If your implementation previously defined the public `iterator.seek(target)`, it must now define the private `iterator._seek(target)`. The new public API is equal to the reference implementation of `leveldown` except for two differences:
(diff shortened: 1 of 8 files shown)


UPGRADING.md
```diff
@@ -153,6 +153,10 @@ suite({
 
 Please see the [README](README.md) for a list of options. Note that some of these have replaced `process.browser` checks.
 
+### Iterator must have a `db` reference
+
+The `db` argument of the `AbstractIterator` constructor became mandatory, as well a public `db` property on the instance. Its existence is not new; the test suite now asserts that your implementation also has it.
```
vweevers: typo: as well \*as a

UPGRADING.md
```diff
@@ -166,7 +170,7 @@ Please see the [README](README.md) for details.
 
 - The default `_clear` method is no longer a noop; instead it clears the operations queued by `_put` and/or `_del`
 - The `_write` method now takes an `options` object as its first argument
-- The `db` argument in the constructor became mandatory, as well the `_db` property on the instance.
+- The `db` argument of the `AbstractChainedBatch` constructor became mandatory, as well a public `db` property on the instance, which was previously named `_db`.
```
vweevers: typo

vweevers:
Will add a changelog entry.

Issue #236, opened 2018-06-16
Proposal: db#clear


vweevers:
Background: https://github.com/Level/abstract-leveldown/pull/24#issuecomment-397760287.

- The default implementation of `clear()` iterates and deletes all keys
- We could add that same fallback to `levelup`, but I don't think we should. Prefer to throw.
- It can take range options, e.g. `clear({ gt: 'foo' })`. Note: it should serialize the range options.
- This allows `subleveldown` to easily implement `clear` on a sublevel
- `deferred-leveldown` should defer it (though it wouldn't need to for the iterator-based implementation)
- `encoding-down` can call `this.db.clear()`. Note: it should encode the range options.
- `memdown` can just create a new tree, unless range options are passed
- `leveldown` can implement it natively (at a later time) for better perf
- `level-js` can call either `IDBObjectStore#clear()` or `IDBObjectStore#delete(range)`

ralphtheninja:
Very nice. I especially like the subleveldown feature, like emptying a bucket of stuff.

vweevers:
Relevant `leveldown` issue: https://github.com/Level/leveldown/issues/9.

vweevers:
Ah, the `level-concat-iterator` dependency comes in handy for the default implementation of `_clear` :)

ralphtheninja:
> Ah, the level-concat-iterator dependency comes in handy for the default implementation of _clear :)

Good. Then we don't have to feel so bad about including it :wink:

vweevers:
> Ah, the level-concat-iterator dependency comes in handy for the default implementation of _clear :)

Scratch that. Would choke on many keys.

vweevers:
Because `abstract-leveldown@6.1.0` now has a `clear()` method, I'm moving this issue to `community` to track overall progress.

PR #310, opened 2018-10-19
Add clear()

vweevers:
Start of #236. For early feedback (see comments in code).

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -183,6 +183,52 @@ AbstractLevelDOWN.prototype._batch = function (array, options, callback) {
   process.nextTick(callback)
 }
 
+AbstractLevelDOWN.prototype.clear = function (options, callback) {
+  if (typeof options === 'function') {
+    callback = options
+  } else if (typeof callback !== 'function') {
+    throw new Error('clear() requires a callback argument')
+  }
+
+  options = cleanRangeOptions(this, options)
+  options.reverse = !!options.reverse
+  options.limit = 'limit' in options ? options.limit : -1
+
+  this._clear(options, callback)
+}
+
+AbstractLevelDOWN.prototype._clear = function (options, callback) {
+  // Avoid setupIteratorOptions, would serialize range options a second time.
+  options.keys = true
+  options.values = false
+  options.keyAsBuffer = true
+  options.valueAsBuffer = true
+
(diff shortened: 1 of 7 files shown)


abstract-leveldown.js
```diff
@@ -179,6 +179,62 @@ AbstractLevelDOWN.prototype._batch = function (array, options, callback) {
...
+
+  // Because the returned records are the same, reverse only makes sense in
+  // combination with limit (e.g. to remove the last N records).
+  options.reverse = !!options.reverse
+  options.limit = 'limit' in options ? options.limit : -1
+
+  this._clear(options, callback)
+}
+
+AbstractLevelDOWN.prototype._clear = function (options, callback) {
```
vweevers: Given the issues described below, perhaps the default `_clear` should just be a noop.

ralphtheninja: Makes sense. Not sure which one though.

vweevers: There's no precedent for throwing, but a noop is undesirable.. unless we enforce the presence of `clear` through tests.

I guess the question is: do we aim to implement `clear` everywhere? Or do we start with partial support, in which case throwing makes more sense.

vweevers: Actually, I prefer putting this feature on hold. V6 already has lots of breaking changes. We have a release dance ahead.

vweevers:
@ralphtheninja Updated. Could you have a look? Tests are yet to do, but I'd first like a review of the general approach, taking into account https://github.com/Level/memdown/pull/191 and https://github.com/Level/level-js/issues/174.

test/common.js
```diff
@@ -19,7 +31,16 @@ function testCommon (options) {
...
     snapshots: options.snapshots !== false,
-    seek: options.seek !== false
+    seek: options.seek !== false,
+    clear: clear
+  }
+}
+
+function warn (msg) {
+  if (typeof process !== 'undefined' && process && process.emitWarning) {
+    process.emitWarning(msg)
```
ralphtheninja: TIL!

vweevers:
Done. Tested against https://github.com/Level/memdown/pull/191.

vweevers:
Once https://github.com/Level/memdown/pull/191 lands in master I'll make a canary branch there to run browser tests as well.

PR #314, opened 2018-11-19
Test deleting with a numeric key

plainweaver:
### `db.get(key[, options][, callback])`

The key can be of any type.


vweevers:
Thanks for contributing and welcome :) but this is already covered by:

https://github.com/Level/abstract-leveldown/blob/af123487c01dc29419b95d1b06445b2ef81e8c04/test/put-get-del-test.js#L117

vweevers:
> wondering why they are separated into put-test, get-test, del-test.

Those test each method individually; their arguments, serialization, asynchronicity, etc. And the `put-get-del-test` is kinda like an "integration test" between them.

plainweaver:
I see... So how the arguments exactly look alike is not their responsibility for each small tests.

It seems intuitive enough.

Issue #248, opened 2018-07-06
Should level-concat-iterator be a dependency?


vweevers:
Otherwise implementations that use the default `testCommon` must include `level-concat-iterator` as a devDependency themselves.

https://github.com/Level/abstract-leveldown/blob/029f56acb2123944045430c54a2fdbdfb69da8d4/package.json#L19-L21

vweevers:
If not, we should at least document these implicit peer dependencies.

PR #326, opened 2019-01-12
Move 'level-concat-iterator' to devDependencies

peakji:
I'm also checking the necessity of dependencies in other level repos.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
     "prepublishOnly": "npm run dependency-check"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


vweevers:
IIRC there was a reason we couldn't do this. Searching GH history..

vweevers:
Yeah: https://github.com/Level/abstract-leveldown/issues/248#issue-338880769

We use `level-concat-iterator` in the abstract test suite, so it must be included as a `dependency` here.

PR #337, opened 2019-04-26
Fix and test asynchronicity of empty batch

### LINKED ISSUE #336: Return early and call back asynchronously if batch is empty
Follow-up for https://github.com/Level/leveldown/pull/619.

vweevers:
Closes #336.

- `db.batch([], callback)` now calls `callback` asynchronously, bypassing `_batch()`
- I considered doing the same for chained batch, by tracking if the batch has operations, but decided not to, so that implementations can choose to ignore certain operations. For example, `batch.del(0)` might be ignored and `abstract-leveldown` would wrongly think there's a queued operation.
- Added abstract tests for asynchronicity of `db.batch([])` (array-form) and `db.batch()` (chained-form)

### DIFF
--- a/test/batch-test.js
+++ b/test/batch-test.js
@@ -193,16 +193,21 @@ exports.args = function (test, testCommon) {
       async = true
     })
   })
+
+  test('test batch() with empty array', function (t) {
+    var async = false
+
+    db.batch([], function (err) {
+      t.error(err, 'no error from batch()')
+      t.ok(async, 'callback is asynchronous')
+      t.end()
+    })
+
+    async = true
+  })
 }
 
 exports.batch = function (test, testCommon) {
-  test('test batch() with empty array', function (t) {
-    db.batch([], function (err) {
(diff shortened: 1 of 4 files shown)


test/batch-test.js
```diff
@@ -193,16 +193,21 @@ exports.args = function (test, testCommon) {
       async = true
     })
   })
-}
 
-exports.batch = function (test, testCommon) {
   test('test batch() with empty array', function (t) {
+    var async = false
```
peakji: <del>Should we change the name `async` into something else since its a reserved keyword in ES8?</del>

peakji: My bad, `async` is not a reserved keyword: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords).

vweevers:
Canary-tested against `memdown` and `level-js` as well, all good.

Issue #235, opened 2018-06-16
Add Symbol.asyncIterator to iterator


vweevers:
Background and use cases in Level/levelup#477 and Level/levelup#491.

MeirionHughes:
I've added an alternative take on this. https://github.com/Level/abstract-leveldown/tree/feature/async-iterator in theory it will just not do anything on older node, but can still be parsed as it manually making the async iterator object.

vweevers:
IMO that much boilerplate code is just not worth it and a sign that async iterators hadn't matured yet. Async generators make it so much cleaner and digestible, similar to what `async/await` does for promises.

MeirionHughes:
true; yes - generators are nice, but there is (or at least was) a major short coming with the generator approach if you absolutely need to clean-up: [you don't know if the user breaks out or not.](https://stackoverflow.com/questions/46414415/how-can-i-find-out-if-a-javascript-iterator-terminates-early). 

i.e. 

```
async function *(){
  yield 1; 
  yield 2; 
  yield 3;
}

for await (let item of source()){
   break;
}
```

My solution (and answer to the SO) [is was to hack the `return`](https://www.npmjs.com/package/@async-generators/terminator), which is even more boiler plate than simply rolling your own `{next, return, error}` object.

MeirionHughes:
I actually gave up on using async iterators last year - I've gone to rxjs for everything at work now. Its just more consistent as you absolutely know when something unsubscribes.

MeirionHughes:
You absolutely want to use a generator you could maybe do it like this: 
```js
AbstractIterator.prototype[Symbol.asyncIterator] = function(){
    let it = (async function * generate() {
      try {
        var kv
    
        while ((kv = await this.next()) !== undefined) {
          yield kv
        }
      } finally {
        await this.end()
      }
    })().bind(this);

    it.return = async function(){
      await this.end();
    }.bind(this)
    
    return it;
  } 

```

vweevers:
> there is (or at least was) a major short coming with the generator approach if you absolutely need to clean-up: you don't know if the user breaks out or not..

That's terrible. Good to know, thanks

PR #338, opened 2019-04-27
Proof of concept: asyncIterator

vweevers:
_Do not merge. Ignore failing tests._

This is to find out what we need to support [`Symbol.asyncIterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator) in userland (we cannot do it here because we need async generators to properly handle ending on errors, and async generators are only available in Node 10+ and Chrome).

The answer is that we need to add Promise support to `iterator.next()` and `.end()`. We can do that in the same way as `levelup`: if you don't provide a callback argument, you get back a Promise.

This means there will be three ways to iterate:

1. callbacks
2. `async/await` (a bit awkward)

```js
const it = db.iterator()
const kv = await it.next()

if (!kv) {
  await it.end()
} else {
  const [key, value] = kv
}
```

3. `for await...of`

```js
for await (let [key, value] of db.iterator()) {
  // no need to call end
}
```

cc @chjj @peakji 

---

Previous discussions:

- https://github.com/Level/abstract-leveldown/issues/235
- https://github.com/Level/leveldown/pull/601
- https://github.com/Level/leveldown/issues/627

### DIFF
--- a/abstract-iterator.js
+++ b/abstract-iterator.js
@@ -9,20 +9,30 @@ function AbstractIterator (db) {
 }
 
 AbstractIterator.prototype.next = function (callback) {
+  // In callback mode, we return `this`
+  var ret = this
   var self = this
 
-  if (typeof callback !== 'function') {
+  if (callback === undefined) {
+    ret = new Promise(function (resolve, reject) {
+      callback = function (err, key, value) {
+        if (err) reject(err)
+        else if (key === undefined && value === undefined) resolve()
+        else resolve([key, value])
+      }
+    })
+  } else if (typeof callback !== 'function') {
     throw new Error('next() requires a callback argument')
   }
 
   if (self._ended) {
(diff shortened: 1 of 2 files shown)


vweevers:
Ah. Implementations like `memdown` that don't need an explicit `end()` call can implement `asyncIterator` themselves without an async generator, thus avoiding syntax errors in older runtimes. The `level-js` iterator currently does need an explicit `end()` but that will change with https://github.com/Level/level-js/issues/86.

So we don't have to do it in userland. The strategy is:

- `abstract-leveldown` implements `Symbol.asyncIterator` with an async generator, lazily required (only in node 10+) and excluded from browserify/webpack builds
- `memdown` (and later `level-js`) implements `Symbol.asyncIterator` without async generator

vweevers:
Superseded by #379.

PR #340, opened 2019-06-22
feat: implement asyncIterator

MeirionHughes:
see https://github.com/Level/abstract-leveldown/issues/235#issuecomment-504695413. WIP PR to trigger tests.

### DIFF
--- a/abstract-iterator.js
+++ b/abstract-iterator.js
@@ -50,7 +50,7 @@ AbstractIterator.prototype.seek = function (target) {
   this._seek(target)
 }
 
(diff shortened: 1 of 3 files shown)


MeirionHughes:
wow the browser tests are taking age. :O

vweevers:
@MeirionHughes re: the browser tests getting stuck, I didn't look, but it's most likely due to some browser (like IE) being unable to parse syntax or not having `Symbol` so it never starts the test. The test suite / tooling can't currently detect that case, and will wait forever for the browser to send back test results. FYI

Issue #345, opened 2019-08-14
Flag to disable iterator tests


piranna:
https://github.com/Level/abstract-leveldown/blob/9d107d0b3049104d8b4464bb6d85b34bd37b61f1/abstract-leveldown.js#L182-L184

Default `_batch` implementation is a no-op, forcing you to implement it yourself to make `abstract-level/test` suite in your own AbstractLevel object. I think it would make sense to have a default implementation that just run over the given operations calling the object methods, something like

```javascript
  _batch(operations, options, callback)
  {
    async.each(operations, ({key, type, value}, callback) =>
    {
      if(type === 'put') return this.put(key, value, options, callback)

      this[type](key, options, callback)
    },
    callback)
  }
```

Main issue would be to add rollback support if one of the operations failed, but maybe this can be left as task for the reader since it's a more advanced and probably underlying backend dependent :-) Thoughs? Would you accept a pull-request for that?

vweevers:
It's hard to make a default implementation atomic.

piranna:
I know, that's why I said about leaving the atomic implementation up to the reader :-) Just only provide a "just enough" default one.

vweevers:
IMO if the default can't be atomic, and therefor cannot pass the test suite (I don't remember if this is actually covered by the test suite), we shouldn't have a default.

piranna:
In that case, maybe a good alternative would be to add a flag to disable batch tests, so it's another way to say "this AbstractLevel doesn't support atomic batch operations", what do you think?

Oh, and said that, maybe also this kind of informative test flags could be set as a metadata property of the AbstractLevel object itself that can be queried in runtime too, but this should be addresed better in a diferent issue :-)

vweevers:
> In that case, maybe a good alternative would be to add a flag to disable batch tests, so it's another way to say "this AbstractLevel doesn't support atomic batch operations", what do you think?

I'm not opposed to adding a flag per se (in addition to the [ones](https://github.com/Level/abstract-leveldown#excluding-tests) we already have), but only if there's an `abstract-leveldown` implementation that can't make batches atomic with good reason.

If we were to add a default implementation of `batch()` that isn't atomic, the default value for the flag would have to be false. Which I don't like. The test suite and its default options should test for full compliance.

> Oh, and said that, maybe also this kind of informative test flags could be set as a metadata property of the AbstractLevel object itself that can be queried in runtime too, but this should be addresed better in a diferent issue :-)

[...]

PR #347, opened 2019-08-16
Option to disable `iterator` tests

piranna:
See #345

### DIFF
--- a/test/chained-batch-test.js
+++ b/test/chained-batch-test.js
@@ -266,6 +266,9 @@ exports.batch = function (test, testCommon) {
         .put('foo', 'bar')
         .write(function (err) {
           t.error(err)
+
+          if(!testCommon.iterator) return t.end()
+
           collectEntries(
             db.iterator({ keyAsBuffer: false, valueAsBuffer: false }), function (err, data) {
               t.error(err)
--- a/test/chained-batch-test.js
+++ b/test/chained-batch-test.js
@@ -267,7 +267,7 @@ exports.batch = function (test, testCommon) {
         .write(function (err) {
           t.error(err)
 
-          if(!testCommon.iterator) return t.end()
+          if (!testCommon.iterator) return t.end()
 
           collectEntries(
             db.iterator({ keyAsBuffer: false, valueAsBuffer: false }), function (err, data) {
--- a/test/common.js
+++ b/test/common.js
@@ -18,6 +18,7 @@ function testCommon (options) {
     bufferKeys: options.bufferKeys !== false,
     createIfMissing: options.createIfMissing !== false,
     errorIfExists: options.errorIfExists !== false,
+    iterator: options.iterator !== false,
     snapshots: options.snapshots !== false,
     seek: options.seek !== false
   }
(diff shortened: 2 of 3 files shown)


vweevers:
I'm -1 on this, because IMO iterators and read streams are a fundamental part of Level. Supporting other use cases can be interesting and the Level org has always been open to experimentation. However, straying too far from the core API means going against user expectations.

Although runtime manifests (https://github.com/Level/community/issues/42) could in the future describe the capabilities of an `abstract-leveldown` implementation, I think they should only describe _additional_ features, and not allow core features to be disabled.

@ralphtheninja thoughts?

ralphtheninja:
> I'm -1 on this, because IMO iterators and read streams are a fundamental part of Level.

I agree.

piranna:
As explained in the other issue thread, [my AbstractLevel](https://github.com/piranna/level-SecureStore) hide keys by using hashes and values by using per-entry encryption keys, so it doesn't make sense to have iterators of entries that you can't know what's their key nor access to its value... Other than this, it's fully compliant. I know this is a corner case usage of the AbstractLevel standar API, but don't know of any other way to make it compliant beyond making iterators optional or decrease the security level of my module... :-/ Any idea?

I understand the reasons why iterators could be considered a basic feature of level stores, but from an usage perspective, it's a level higher of what a basic key-value store can do...

vweevers:
> don't know of any other way to make it compliant beyond making iterators optional or decrease the security level of my module... :-/ Any idea?

I'm afraid the conclusion is that it can't be compliant. I am wondering though why you picked `abstract-leveldown` as your interface? If not to integrate with the Level ecosystem.

piranna:
>  am wondering though why you picked abstract-leveldown as your interface? If not to integrate with the Level ecosystem.

Yeah! That's exactly the reason. At first I was thinking about using directly `level` module, but seeing I was able to do it with minor changes in a more standard, compatible and reusable way, I decided to move that way and make it integrable. This way I could focus on the encryption part and later be able to change both the backend database or access it from a REST API.

vweevers:
I'm not sure how much value there is to integrating with the Level ecosystem without iterators. I suggest to clearly document that limitation of your API.

As for the test suite, you could just copy https://github.com/Level/abstract-leveldown/blob/master/test/index.js and take out the iterator tests. Do note that we assume that only `abstract-leveldown/test` is directly required by dependents; we may move internal files around without (semver) notice.

piranna:
> I suggest to clearly document that limitation of your API.

That's was part of my plan, at least for short-mid term.


> As for the test suite, you could just copy [/test/index.js@`master`](https://github.com/Level/abstract-leveldown/blob/master/test/index.js) [](/Level/abstract-leveldown/blob/HEAD@{2019-08-17T20:34:18Z}/test/index.js) and take out the iterator tests. Do note that we assume that only `abstract-leveldown/test` is directly required by dependents; we may move internal files around without (semver) notice.

I'm inheriting from the `AbstractLevel` class, too. Only missing part from the API... the iterators :-P

vweevers:
> I'm inheriting from the AbstractLevel class, too

Right. To clarify, I was only talking about the `test/` directory. You can `require('abstract-leveldown/test')` for the whole suite or `require('abstract-leveldown/test/batch-test')` etc for individual tests, but the latter may be moved.

vweevers:
I hope you have sufficient info to move forward now.

PR #351, opened 2019-09-20
Add manifest

vweevers:
See Level/community#83.

Related work: https://github.com/Level/levelup/pull/674, https://github.com/Level/level-js/pull/179.

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -1,54 +1,16 @@
 var xtend = require('xtend')
+var supports = require('level-supports')
 var AbstractIterator = require('./abstract-iterator')
 var AbstractChainedBatch = require('./abstract-chained-batch')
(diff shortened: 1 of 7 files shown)


abstract-leveldown.js
```diff
@@ -4,8 +4,52 @@ var AbstractChainedBatch = require('./abstract-chained-batch')
...
+      // Example
+      // approximateSize: {
+      //   sync: false,
+      //   callback: true,
+      //   promise: false
+      // },
+      // createWriteStream: {
+      //   sync: true,
+      //   writable: true
+      // }
```
vweevers: Not sure how far to take this (initially).

I'm thinking I'll create a `level-supports` module (under Level), and we'll discuss additions there?

PR #353, opened 2019-10-20
Add more range tests

vweevers:
To try and break https://github.com/Level/leveldown/pull/689/commits/fb7e970ba3dc68d652b0c4be5a038c71c9ed7c70 of https://github.com/Level/leveldown/pull/689.

Tested locally against `leveldown`, `memdown` and `level-js`.

### DIFF
--- a/test/iterator-range-test.js
+++ b/test/iterator-range-test.js
@@ -1,4 +1,5 @@
 var collectEntries = require('level-concat-iterator')
+var xtend = require('xtend')
 
 var db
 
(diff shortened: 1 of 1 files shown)


test/iterator-range-test.js
```diff
@@ -44,6 +45,30 @@ exports.range = function (test, testCommon) {
...
+        reverseOpts.start = opts.end
+        delete reverseOpts.end
+      }
+
+      rangeTest(
+        name + ' (flipped)',
+        reverseOpts,
+        expected.slice().reverse()
+      )
+    }
```
vweevers: This makes some tests (with `reverse=true`) redundant. We can remove those later, I figured that would be easier to review as a separate PR.

PR #355, opened 2019-11-19
src: Import buffer

Raynos:
Do not rely on the global.Buffer property implicitly.

In some environments like `electron` this global variable
does not exist at all.

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -1,5 +1,6 @@
 var xtend = require('xtend')
 var supports = require('level-supports')
+var Buffer = require('buffer').Buffer
 var AbstractIterator = require('./abstract-iterator')
 var AbstractChainedBatch = require('./abstract-chained-batch')
 var hasOwnProperty = Object.prototype.hasOwnProperty


Raynos:
cc @vweevers @heapwolf 

This adds `electron` support to the `abstract-leveldown` package.

vweevers:
Hm, we test `leveldown` and `rocksdb` in Electron and have not encountered this issue. Is that because we run those tests in the main process, rather than renderer processes?

Raynos:
@vweevers I believe that is correct.

This is where in electron it deletes `global.Buffer` ( https://github.com/electron/electron/blame/master/lib/renderer/init.ts#L188-L196 ).

vweevers:
How about doing:

```js
const window = new BrowserWindow({
  webPreferences: {
    nodeIntegration: true
  }
})
```

Raynos:
It also looks like the root cause of my issue is very recent

https://github.com/electron/electron/pull/17838/files

It came in a PR for node12 from April this year.

vweevers:
Checkout https://github.com/Level/electron-demo/blob/master/index.js

Raynos:
@vweevers 

The actual issue is here https://github.com/electron/electron/blame/master/lib/renderer/init.ts#L21

```
Module.wrapper = [
  '(function (exports, require, module, __filename, __dirname, process, global, Buffer) { ' +
  // By running the code in a new closure, it would be possible for the module
  // code to override "process" and "Buffer" with local variables.
  'return function (exports, require, module, __filename, __dirname) { ',
  '\n}.call(this, exports, require, module, __filename, __dirname); });'
]
```

Recent versions of electron evaluate all code in a closure that shadow `global.Buffer` with a function argument called `Buffer` whose value is always `undefined` 

So the `Buffer` reference in the source code is not a global variable, it is function argument whose value is `undefined` ;

I guess I can try to get `electron` to fix this instead but not relying on global variables seems like a good thing anyway, explicit import is better then global variable reference.

vweevers:
So does https://github.com/Level/abstract-leveldown/pull/355#issuecomment-555444354 no longer work?

Raynos:
I can't seem to reproduce the issue in the `electron-demo` repository...

vweevers:
Note, I'm not opposed to `require('buffer')` per se, just worried we'll have to do it in each and every Level module, so if it can be solved in another way, I'd prefer that.

Raynos:
Turns out the root cause was the module `v8-compile-cache` ( https://github.com/zertosh/v8-compile-cache/issues/3 ).

vweevers:
Alright. Is that module used in Electron directly (in which case I'd expect `electron-demo` to break when we update that to latest Electron), or by your app (in which case, there's nothing to do here, right)?

Raynos:
That module is used by my application, so there's nothing to do here. 

I'd still recommend merging this for hygiene.

PR #357, opened 2019-12-19
src: move level-concat-iterator to dev deps

Raynos:
The level-concat-iterator dependency is not used by any
of the production code, it's a utility library used in the test
suite and should be marked as a devDependencies.

cc @vweevers


vweevers:
It's used by the test suite and therefor should be in deps.

Raynos:
Ah; makes sense.

PR #362, opened 2020-03-27
fix: add buffer

hugomrdias:
This PR add buffer as a dependency.

Related to https://github.com/Level/level-js/pull/191

@vweevers regarding process.nextTick should we copy immediate.js and immediate-browser.js from level-js and then remove and require in level-js from here ?


vweevers:
> @vweevers regarding process.nextTick should we copy immediate.js and immediate-browser.js from level-js and then remove and require in level-js from here ?

I prefer not to change the timing behavior in node, because if folks are using `process.nextTick` in their app or module but this switches to `setImmediate` that changes the order of execution. Another but lesser concern is that we'll have to update tests that rely on the order.

vweevers:
Hm, could it be considered a semver-major change in browsers as well?

hugomrdias:
for reference from streams repo https://github.com/nodejs/readable-stream/pull/414

recap:
- keep node with process.nextTick
- browser with https://github.com/calvinmetcalf/immediate#readme (same as level-js)

and no i dont think it would be a breaking change in browsers nextTick injected by bundlers is mostly what `immediate` does.

hugomrdias:
for reference browserify and webpack use this https://github.com/defunctzombie/node-process/blob/master/browser.js#L134

hugomrdias:
can i go forward and update this PR with this ?

> recap:
> 
> keep node with process.nextTick
> browser with https://github.com/calvinmetcalf/immediate#readme (same as level-js)

vweevers:
They (immediate and the process polyfill) seem similar enough, yeah. Both have microtask behavior and a draining queue.

vweevers:
Possible issue: `immediate` doesn't work in web workers. See https://github.com/calvinmetcalf/immediate/issues/27 and https://github.com/pouchdb/pouchdb/issues/7973. Apparently this is already a problem for `level-js` and `memdown` today, so this is not a blocker, but I did want to make note of it.

hugomrdias:
try my runner to test in the browser https://github.com/hugomrdias/playwright-test i have had serious problems with airtap not bundling correctly and the DX is much nicer. `playrwright-test test/self.js --runner tape`

vweevers:
I've been working on adding playwright to airtap actually 😄 (and any other browser for that matter).

hugomrdias:
> I've been working on adding playwright to airtap actually 😄 (and any other browser for that matter).

cool, any feedback would be awesome or if you need any help please ping me.

vweevers:
6.2.3

Raynos:
This increased the dependency count for `leveldown` :(

vweevers:
Yeah, it's hard to satisfy everyone's needs. Ideas?

Raynos:
I've got an idea for `leveldown` that I am making a PR for. It will not impact `abstract-leveldown` ;

Raynos:
Ok opened https://github.com/Level/leveldown/pull/714

hugomrdias:
> I've been working on adding playwright to airtap actually 😄 (and any other browser for that matter).
> 
> What does DX mean?

Developer Experience

hugomrdias:
> Possible issue: `immediate` doesn't work in web workers. See [calvinmetcalf/immediate#27](https://github.com/calvinmetcalf/immediate/issues/27) and [pouchdb/pouchdb#7973](https://github.com/pouchdb/pouchdb/issues/7973). Apparently this is already a problem for `level-js` and `memdown` today, so this is not a blocker, but I did want to make note of it.

this seems to be related only to the use of `global` inside immediate

vweevers:
@hugomrdias  Would you mind opening an issue in airtap to describe your experience? Unless it's about webpack, then the answer is simply "yeah, we're browserify die-hards, and you _should_ be using another tool".

hugomrdias:
> @hugomrdias Would you mind opening an issue in airtap to describe your experience? Unless it's about webpack, then the answer is simply "yeah, we're browserify die-hards, and you _should_ be using another tool".

sure, and it’s not about webpack actually i don't care about webpack (i want to add support for more bundlers in the same way it already support multiple test runners) its about giving the developer the same experience that they have in node (everything is properly redirected to the terminal) and if they really need to run in debug mode (browser window opened) they always have a clean env, devtools opened already focused in the console (playwright needs to land support for this already talked with them and they support it). Also it supports extensions env, webworkers etc

hugomrdias:
@vweevers should we change to https://github.com/YuzuJS/setImmediate to fix the `global` issue in webworkers ?

vweevers:
Thanks for the airtap feedback! Agree on all points, and there's an effort underway to do all that except for extensions and webworkers (although that would be easier to implement in the next version).

> @vweevers should we change to https://github.com/YuzuJS/setImmediate to fix the `global` issue in webworkers ?

That's a macrotask library, while `immediate` is a microtask library (although they use some of the same tricks, so uuuhm). I prefer the latter, to stay as close as possible to `process.nextTick()` behavior.

Raynos:
> This increased the dependency count for leveldown :(

Opened two PRs to reduce dependencies

 - https://github.com/Level/abstract-leveldown/pull/365
 - https://github.com/Level/abstract-leveldown/pull/366

vweevers:
@Raynos A heads up: I'm swamped atm

Raynos:
No worries :)

hugomrdias:
ofc .. my bad what about https://github.com/feross/queue-microtask (https://nodejs.org/api/globals.html#globals_queuemicrotask_callback)

vweevers:
@hugomrdias re: immediate and web workers, I suggest that whoever needs that sends a PR to `immediate`. Ideally we all rally behind a single module, to avoid ending up with multiple modules in `node_modules` that all more or less do the same thing.

PR #364, opened 2020-04-05
Support running test suite on a levelup db

vweevers:
As well as skipping 'start' and 'end' tests (for `multileveldown`) (see also https://github.com/Level/community/issues/86).

Doesn't change anything for existing `abstract-leveldown` implementations. Before releasing I do want to test it on:

- [x] `multileveldown` (as a `levelup` instance)
- [x] `subleveldown` (same)
- [x] `level`

Working towards https://github.com/Level/community/issues/58.

### DIFF
--- a/test/batch-test.js
+++ b/test/batch-test.js
@@ -11,7 +11,7 @@ exports.setUp = function (test, testCommon) {
 }
 
 exports.args = function (test, testCommon) {
-  test('test callback-less, 2-arg, batch() throws', function (t) {
+  testCommon.promises || test('test callback-less, 2-arg, batch() throws', function (t) {
     t.throws(
       db.batch.bind(db, 'foo', {}),
       /Error: batch\(array\) requires a callback argument/,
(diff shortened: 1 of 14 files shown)


test/batch-test.js
```diff
@@ -131,7 +131,8 @@ exports.args = function (test, testCommon) {
     async = true
   })
 
-  test('test batch() with missing array', function (t) {
+  // TODO: align abstract-leveldown with levelup or vice versa
+  testCommon.levelup || test('test batch() with missing array', function (t) {
```
vweevers, line 135: Should fix these inconsistencies now, because otherwise we'll have to make semver-major changes later.

vweevers:
The `createIfMissing` and `errorIfExists` tests can't be fixed without a semver-major change, because they do:

```js
test('test database open createIfMissing:false', function (t) {
  // levelup db, opens itself
  var db = testCommon.factory()

  // options are ignored because it's already opening
  db.open({ createIfMissing: false }, function () { /* .. */ })
})
```

The easiest way forward is probably to move such options to the `(abstract)-leveldown` constructor.

PR #366, opened 2020-04-27
tests: Move level-concat-iterator to optional

Raynos:
This means that it's only required for testing and not
a production dependency.

I've added a human readable error with instructions on how
to fix it for the users of this package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,9 +19,11 @@
   "dependencies": {
     "buffer": "^5.5.0",
     "immediate": "^3.2.3",
-    "level-concat-iterator": "~2.0.0",
     "level-supports": "~1.0.0",
     "xtend": "~4.0.0"
+  },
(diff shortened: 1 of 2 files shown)


vweevers:
I'm not a fan of optionalDependencies. They're ambiguous, and npm ignores installation failures.

I think I would prefer moving the test suite to its own package (not sure how that should be versioned). Also because I've recently been bringing the `abstract-leveldown` and `levelup` test suites & interfaces closer together (there are now `abstract-leveldown` implementations that call the `levelup` test suite, and vice versa), with the ultimate goal of having one test suite.

Raynos:
Cool.

Raynos:
I mean; thanks for taking a look. It’s hard to maintain something used by so many modules. 

We appreciate the maintenance.

PR #371, opened 2020-12-02
Added code to run on Power systems

genisysram:
Hi
Here is my contribution to your code, its working good on powersystems.

Thanks for the code, its working good.
## What do these changes do?
Added Architecture "ppc64le"
## Are there changes in behavior for the user?
No

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,21 @@ matrix:
     env: CMD=test
   - node_js: stable
     env: CMD=test-browsers
+#adding code to work on Power systems
+  fast_finish: true
+  include:
+  - node_js: 6
+    env: CMD=test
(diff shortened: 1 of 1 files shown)


vweevers:
Thanks for the PR. My response is the same as https://github.com/browserify/node-util/pull/53#issuecomment-744564852:

> Nothing about this package (or any pure-JS package that doesn't touch the filesystem) can vary based on architecture, so testing in other architectures is not valuable.

On `abstract-leveldown` we can guarantee we'll never introduce dependencies with native addons, or touch the filesystem, because `abstract-leveldown` is an abstract interface that targets both node and browsers.

Testing against ppc64le would be valuable on native addons like https://github.com/Level/leveldown.

PR #373, opened 2021-03-31
chore: update deps

achingbrain:
Updates `Buffer` to the latest version as having multiple versions in
the dep tree causes quite a lot of bundle bloat when run in browsers!

### DIFF
--- a/.airtap.yml
+++ b/.airtap.yml
@@ -2,8 +2,6 @@ browsers:
   - name: chrome
     version: latest
     platform: Windows 10
-  - name: internet explorer
-    version: '11'
   - name: firefox
     version: latest
     platform: Windows 10
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "prepublishOnly": "npm run dependency-check"
   },
   "dependencies": {
-    "buffer": "^5.5.0",
+    "buffer": "^6.0.3",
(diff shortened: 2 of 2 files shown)


vweevers:
Semver-major because it drops support of IE11. @achingbrain Could you remove IE11 from `.airtap.yml`?

achingbrain:
Done!

vweevers:
Ran browser tests on a temporarily expanded matrix, all good! Safari 11 on Mac failed, but seeing as iOS Safari 11 did succeed, I think it was a random or at least unrelated failure. Will retry that.

image

vweevers:
Was indeed unrelated. Safari threw `TypeError: Value is not a sequence` upon `console.log(..)`, which is caused by [the WebDriver extension that Sauce Labs has installed](https://github.com/apache/cordova-plugin-media-capture/issues/105#issuecomment-425052111) on MacOS 10.12, but luckily not on 10.13 so I switched to that. Tests passed.

vweevers:
I've since removed the need for `buffer`: https://github.com/Level/abstract-leveldown/commit/8f792349bb940bac47152c7f02fae92c081afac7

vweevers:
[7.0.0](https://github.com/Level/abstract-leveldown/releases/tag/v7.0.0)

PR #374, opened 2021-04-03
Drop legacy features & runtime environments

vweevers:
This PR drops support of:

- Legacy range options: `start` and `end` (https://github.com/Level/community/issues/86)
- Node 6 and 8
- IE11 and Safari 9-11 (also because of #373)
- Stock Android browser (AOSP)

This allows us to remove `xtend`, use latest dependencies (`nyc`, `sinon`, `standard`) which in turn removes vulnerabilities, and to use async generators (in a future semver-minor PR). Easiest to review commit by commit; do not squash on merge.

Ref https://github.com/Level/community/issues/98.

FYI @achingbrain because this slightly delays the release of your #373.

### DIFF
--- a/abstract-iterator.js
+++ b/abstract-iterator.js
@@ -1,7 +1,5 @@
 'use strict'
 
-const nextTick = require('./next-tick')
-
 function AbstractIterator (db) {
   if (typeof db !== 'object' || db === null) {
     throw new TypeError('First argument must be an abstract-leveldown compliant store')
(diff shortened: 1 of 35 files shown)


abstract-iterator.js
```diff
@@ -11,7 +13,7 @@ function AbstractIterator (db) {
 }
 
 AbstractIterator.prototype.next = function (callback) {
-  var self = this
+  const self = this
```
ralphtheninja, line 16: Maybe get rid of these `self` variables completely?

ralphtheninja (approved):
Just a minor note on `self` variables.

vweevers:
[7.0.0](https://github.com/Level/abstract-leveldown/releases/tag/v7.0.0)

PR #375, opened 2021-04-20
chore(ci): converted travis workflow to github action

marcuslyons:
Ref [Level/community#99](https://github.com/Level/community/issues/99)


.github/workflows/sauce.yml
```diff
@@ -0,0 +1,25 @@
+name: Sauce Labs
+on: push
+concurrency: sauce-labs
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
```
vweevers, line 9:
```suggestion
        uses: actions/checkout@v2
```

The action we copied from was a bit outdated

.github/workflows/sauce.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Set up node
+        uses: actions/setup-node@v1
```
vweevers, line 11:
```suggestion
        uses: actions/setup-node@v2
```

marcuslyons: Changed in b660785

.github/workflows/sauce.yml
marcuslyons: Changed in b660785

PR #376, opened 2021-05-06
feat: map method with default _map fallback

MeirionHughes:
As per https://github.com/Level/community/issues/100

Adds new `map` method to serve as a "multi-get" / "batch-get" method for implementing 'downs. 

Adds protected overload-able `_map` method with default fallback to looping over each key and using `_get`. i.e. non-breaking for non-implementing 'downs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,14 @@ Store a new entry or overwrite an existing entry. There are no `options` by defa
 
 Delete an entry. There are no `options` by default but implementations may add theirs. The `callback` function will be called with no arguments if the operation is successful or with an `Error` if deletion failed for any reason.
 
+### `db.map(keys[, options], callback)`
+
+Get multiple values from the store by an array `keys`. The optional `options` object may contain:
+
+- `asBuffer` _(boolean, default: `true`)_: Whether to return each `value` in the result array as a Buffer. If `false`, the returned type depends on the implementation.
+
+The `callback` function will be called with an `Error` if the operation failed for any reason. If successful the first argument will be `null` and the second argument will be the array of values. Given N keys, the result will contain N values.
+
 ### `db.batch(operations[, options], callback)`
 
 Perform multiple _put_ and/or _del_ operations in bulk. The `operations` argument must be an `Array` containing a list of operations to be executed sequentially, although as a whole they are performed as an atomic operation.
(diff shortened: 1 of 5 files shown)


MeirionHughes:
this needs more tests to confirm _map is working properly

MeirionHughes:
seems not all tests are ran within abstract-down's test script - I've added a test into self.js now to check the default loop-get fallback. Not sure how to make sure the `export.map` tests are okay though.

README.md
```diff
@@ -427,6 +435,14 @@ Delete an entry. There are no default options but `options` will always be an ob
 
 The default `_del()` invokes `callback` on a next tick. It must be overridden.
 
+### `db._map(keys, options, callback)`
+
+Map each `key` in the `keys` array to a `value`. The `options` object will always have the following properties: `asBuffer`. If _any_ key in `keys` does not exist, call the `callback` function with a `new Error('NotFound')`. Otherwise call `callback` with `null` as the first argument and the values array as the second.
```
vweevers, line 440: > If _any_ key in `keys` does not exist, call the `callback` function with a `new Error('NotFound')`

Not sure this is desirable. You wouldn't know which key wasn't found. With parallel writes, it's not unlikely that `keys` will contain deleted keys.

abstract-leveldown.js
```diff
@@ -133,6 +133,59 @@ AbstractLevelDOWN.prototype._del = function (key, options, callback) {
...
 }
 
+AbstractLevelDOWN.prototype.map = function (keys, options, callback) {
+  if (typeof options === 'function') callback = options
+
+  if (typeof callback !== 'function') {
+    throw new Error('map() requires a callback argument')
+  }
+
+  if (Array.isArray(keys) === false || keys === null) {
```
vweevers, line 143:
```suggestion
  if (!Array.isArray(keys)) {
```

The null-check is redundant.

abstract-leveldown.js
```diff
@@ -133,6 +133,59 @@ AbstractLevelDOWN.prototype._del = function (key, options, callback) {
...
+    return this._nextTick(callback, new Error('map(keys) element must be an array and not `null`'))
+  }
+
+  const _keys = new Array(keys.length)
+
+  for (const index in keys) {
```
vweevers, line 149: MDN [discourages `for..in` on arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#array_iteration_and_for...in).

abstract-leveldown.js
```diff
@@ -133,6 +133,59 @@ AbstractLevelDOWN.prototype._del = function (key, options, callback) {
...
+
+  this._map(_keys, options, callback)
+}
+
+AbstractLevelDOWN.prototype._map = function (keys, options, callback) {
+  const count = keys.length
+  const result = new Array(count)
+  let index = 0
+
+  if (count === 0) return this._nextTick(callback, null, [])
```
vweevers, line 169: Can we do this in `map()` instead?

README.md
MeirionHughes: I'll remove the default fallback - but how would you want to deal with this? if it fails on the n'th item, then return an array with n-1 items _and_ a non-null error?

vweevers: I was thinking that we could yield an `undefined` value. For example, given keys `['a', 'b', 'c']` and key `b` does not exist, the result would be `[value, undefined, value]`.

vweevers:
Superseded by https://github.com/Level/abstract-leveldown/pull/381

PR #379, opened 2021-09-12
Support `for await...of db.iterator()`

### LINKED ISSUE #235: Add Symbol.asyncIterator to iterator

vweevers:
Closes #235. You can now do:

```js
try {
  for await (const [key, value] of db.iterator()) {
    console.log(key)
  }
} catch (err) {
  console.error(err)
}
```

Supersedes #338, which was just a proof of concept and we've since dropped support of legacy runtimes (Level/community#98) which now allows us to use async generators across the board.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,31 +311,53 @@ A reference to the `db` that created this chained batch.
 
 An iterator allows you to _iterate_ the entire store or a range. It operates on a snapshot of the store, created at the time `db.iterator()` was called. This means reads on the iterator are unaffected by simultaneous writes. Most but not all implementations can offer this guarantee.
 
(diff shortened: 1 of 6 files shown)


ralphtheninja (approved):

vweevers:
Might want to add `keys` and `values` options before it's too late to change the function signature. We could do:

```js
for await (const key of db.iterator({ values: false }))
```

But I think I prefer a new API, in which case it's not a blocker:

```js
for await (const key of db.keyIterator())
```

Or even:

```js
for await (const key of db.keys())
```

ralphtheninja:
I prefer the last one.

vweevers:
[7.1.0](https://github.com/Level/abstract-leveldown/releases/tag/v7.1.0)

PR #381, opened 2021-09-25
Add `db.getMany(keys)`

vweevers:
Tested against `leveldown`, `memdown`, `level-js`, `deferred-leveldown`, `encoding-down`, (indirectly) `levelup`, `subleveldown`.

Ref https://github.com/Level/community/issues/101

### DIFF
--- a/abstract-leveldown.js
+++ b/abstract-leveldown.js
@@ -364,13 +364,14 @@ AbstractLevelDOWN.prototype._checkValue = function (value) {
   }
 }
 
-// Undocumented, only here for levelup API parity
-AbstractLevelDOWN.prototype.isOpen = function () {
-  return this.status === 'open'
+// TODO: docs and tests
+AbstractLevelDOWN.prototype.isOperational = function () {
+  return this.status === 'open' || this._isOperational()
 }
 
-AbstractLevelDOWN.prototype.isClosed = function () {
-  return this.status === 'new' || this.status === 'closing' || this.status === 'closed'
+// Implementation may accept operations in other states too
+AbstractLevelDOWN.prototype._isOperational = function () {
+  return false
 }
 
 // Expose browser-compatible nextTick for dependents
@@ -381,20 +382,7 @@ AbstractLevelDOWN.prototype._nextTick = require('./next-tick')
 module.exports = AbstractLevelDOWN
 
 function maybeError (db, callback) {
-  if (db.status !== 'open') {
-    // Exemption to support deferredOpen
-    if (db.type === 'deferred-leveldown') {
-      return false
-    }
-
-    // TODO: deferred-leveldown and levelup are inconsistent: the former allows
-    // operations on any db status, the latter only if 'opening' or 'open'. We
-    // should define the scope of the deferredOpen feature. For now, pick the
-    // levelup behavior. Ref https://github.com/Level/community/issues/58
-    if (db.supports.deferredOpen && db.status === 'opening') {
-      return false
-    }
-
+  if (!db.isOperational()) {
     db._nextTick(callback, new Error('Database is not open'))
     return true
   }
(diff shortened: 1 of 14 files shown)


abstract-leveldown.js
```diff
@@ -90,6 +94,51 @@ AbstractLevelDOWN.prototype._get = function (key, options, callback) {
...
+      this._nextTick(callback, err)
+      return callback.promise
+    }
+
+    serialized[i] = this._serializeKey(key)
+  }
+
+  this._getMany(serialized, options, callback)
+  return callback.promise
+}
```
vweevers, line 136: Note that here we're doing everything that levelup normally does, as an experiment.

abstract-leveldown.js
```diff
@@ -315,9 +364,40 @@ AbstractLevelDOWN.prototype._checkValue = function (value) {
...
+      return false
+    }
+
+    // TODO: deferred-leveldown and levelup are inconsistent: the former allows
+    // operations on any db status, the latter only if 'opening' or 'open'. We
+    // should define the scope of the deferredOpen feature. For now, pick the
+    // levelup behavior. Ref https://github.com/Level/community/issues/58
+    if (db.supports.deferredOpen && db.status === 'opening') {
+      return false
+    }
```
vweevers, line 396: Not pretty but it works

abstract-leveldown.js
```diff
@@ -90,6 +94,51 @@ AbstractLevelDOWN.prototype._get = function (key, options, callback) {
...
+  callback = getCallback(options, callback)
+  callback = catering.fromCallback(callback)
+  options = getOptions(options)
+
+  if (maybeError(this, callback)) {
+    return callback.promise
+  }
+
+  if (!Array.isArray(keys)) {
+    this._nextTick(callback, new Error('getMany() requires an array argument'))
```
vweevers, line 107: I copied this async behavior from `levelup#batch`. Can't remember why we're not throwing there. I opted to be consistent with `levelup` just to make a choice, in the grand scheme of things this is not important.

test/get-many-test.js
```diff
@@ -0,0 +1,240 @@
...
+  })
+}
+
+exports.getMany = function (test, testCommon) {
+  test('test getMany() support is reflected in manifest', function (t) {
+    t.is(db.supports && db.supports.getMany, true)
+    t.end()
+  })
+
+  test('test simple getMany()', function (t) {
```
vweevers, line 65: I copied this from `get-test.js`

abstract-leveldown.js
vweevers: Which means that `levelup` only has to do [this](https://github.com/Level/levelup/blob/a837e5407f2da39924ed327a655521fefcb0a214/lib/levelup.js#L188-L190):

```js
LevelUP.prototype.getMany = function (keys, options, callback) {
  return this.db.getMany(keys, options, callback)
}
```

abstract-leveldown.js
vweevers: It makes sense when considering `deferred-leveldown`, would otherwise throw once the db is open.

vweevers:
Added in the last commit (1f1ab97):

- `AbstractLevelDOWN.prototype.isOperational()` which returns true if the db accepts operations. Normally that means when `db.status` is `'open'`, and `deferred-leveldown` can override `_isOperational()` to also accept operations when `db.status` is `'opening'`.
- More tests

And removed:

- `test custom _serialize*` tests which were not that valuable and not monkeypatch-safe

abstract-leveldown.js
ralphtheninja: So later on we can move over the `catering` stuff to `abstract-leveldown` one by one.

ralphtheninja (approved):

PR #382, opened 2021-10-02
Add `length` property to chained batch for API parity

vweevers:
Same as https://github.com/Level/levelup/pull/372.


vweevers:
Now a part of https://github.com/Level/abstract-leveldown/pull/383.

PR #383, opened 2021-10-03
Increase API parity with `levelup`

vweevers:
Biting the bullet. This adds promise support, makes `open()` & `close()` idempotent, adds open and close events, makes an `abstract-leveldown` db safer than a `levelup` db, and more.

Needs a ton of (canary) tests. Contains a few subtle breaking changes, but for consumers of downstream implementations it should be painless. An `abstract-leveldown` db can still be wrapped with `levelup` and folks can continue to use that interface.

I'm opening this up for early review, I might take it even further.

Ref https://github.com/Level/community/issues/58
Ref https://github.com/Level/leveldown/issues/8


vweevers:
Complexity is slowly increasing here and I'm wary of that. So far it's still worth it. I find the behavior easier to oversee, and reimplementing certain `levelup` features (like idempotent open) gives me a chance to polish them and remove the need for safety checks that we've had to add to `leveldown`.

I've been canary testing this against various modules; `subleveldown` in particular because I sense that this module will tell me when I've found a balance between increased `abstract-leveldown` complexity and reduced `subleveldown` complexity.

vweevers:
Funny how this is starting to look like the `levelup` from [8 years ago](https://github.com/Level/levelup/blob/3f3fc0a76afd11bf3ac4f3020e850271913a19fd/lib/levelup.js).

> An `abstract-leveldown` db can still be wrapped with `levelup` and folks can continue to use that interface.

Technically that still works (with a few tweaks in `levelup`) but there's no value in doing that anymore (except if you need streams, but that will be easy to [move to a module](https://github.com/Level/community/discussions/105)). Thinking about forking `abstract-leveldown` including dependents, for a clean break. For now I'm continuing here, because I have no reason to publish it yet and I've yet to prove this is all worth it. Next up are encodings and maybe a light form of hooks.

vweevers:
Moved to https://github.com/Level/abstract-level.

